PR #4, opened 2021-06-28
Add a note in the README about how to get .js files from Github.

sarahcstringer:
This adds a section for how to pull in the sdk code from github. Because the CDN is no longer available, we want to show an option for how to get the code without node or npm (for example, if someone is just testing the sdk with a server in a different language, and doesn't want to download node to run this).

**Contributing to Twilio**

### Before merge
* [ ] Got one or more +1s


sarahcstringer:
Closing in favor of #5

PR #6, opened 2021-06-30
Remove `Device.activeCall`.

mhuynh5757:
**Contributing to Twilio**

### JIRA link(s):

- [VBLOCKS-145](https://issues.corp.twilio.com/browse/VBLOCKS-145)

### Description

Remove `Device.activeCall` public class member.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Breaking API Changes
 
 ### Active call no longer accessible through the Device object.
 
-To future-proof the SDK, `Device.activeCall` is no longer available. Instead,
+`Device.activeCall` is no longer available. Instead,
 the application should keep track of active calls itself, i.e. when accepting an
 incoming call or when making an outgoing call.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,32 @@
+2.0.0 (In Progress) - Release
+===================
+
+Breaking API Changes
+--------------------
+
+### Active call no longer accessible through the Device object.
+
+To future-proof the SDK, `Device.activeCall` is no longer available. Instead,
+the application should keep track of active calls itself, i.e. when accepting an
+incoming call or when making an outgoing call.
+
+```ts
+const token = '...';
+const options = { ... };
+const device = new Device(token, options);
+
+await device.register();
+
+const activeCall = await device.connect();
+```
+
 2.0.0 (Apr 30, 2021) - Pilot
 ===================
 
 #### This version is currently in the Pilot phase!
 This product, Twilio's JavaScript Voice SDK, is the next version of Twilio's Javascript Client SDK. It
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,10 +1,32 @@
+2.0.0 (In Progress) - Release
+===================
+
+Breaking API Changes
+--------------------
+
+### Active call no longer accessible through the Device object.
+
+To future-proof the SDK, `Device.activeCall` is no longer available. Instead,
```
charliesantos, line 9: Do we really need to mention "To future-proof the SDK"? I think we can remove this, it doesn't tell me why this is future proofing as the customer.

charliesantos (approved):

CHANGELOG.md
```diff
@@ -1,10 +1,32 @@
...
+const token = '...';
+const options = { ... };
+const device = new Device(token, options);
+
+await device.register();
+
+const activeCall = await device.connect();
+```
+
 2.0.0 (Apr 30, 2021) - Pilot
```
idelgado, line 23: I would expect a Pilot release to have a specifier at the end to indicate as much. Did it have a preview or beta suffix?

CHANGELOG.md
```diff
@@ -1,10 +1,32 @@
+2.0.0 (In Progress) - Release
+===================
+
+Breaking API Changes
+--------------------
+
+### Active call no longer accessible through the Device object.
+
+`Device.activeCall` is no longer available. Instead,
+the application should keep track of active calls itself, i.e. when accepting an
```
idelgado, line 10: > Instead, the application should keep a reference to any call that is made using connect or accepted.

CHANGELOG.md
```diff
@@ -1,10 +1,32 @@
...
+incoming call or when making an outgoing call.
+
+```ts
+const token = '...';
+const options = { ... };
+const device = new Device(token, options);
+
+await device.register();
+
+const activeCall = await device.connect();
```
idelgado, line 20: I would just name this `call`.

CHANGELOG.md
```diff
@@ -1,10 +1,32 @@
+2.0.0 (In Progress) - Release
```
idelgado, line 1: When 2.0.0 is released will the changelog for the Pilot and the Release be combined? As a developer upgrading to 2.0.0 from 1.X I would expect all the changes to be in one place.

CHANGELOG.md
mhuynh5757: It did not, but it was also never released to `npm`, only a GitHub release tag.

CHANGELOG.md
mhuynh5757: I believe this should be the case, I will talk to Brandon Hawkins about this.

PR #10, opened 2021-07-06
[VBLOCKS-146] Fix double register during `Device.updateOptions`.

mhuynh5757:
**Contributing to Twilio**

### JIRA link(s):

- [VBLOCKS-146](https://issues.corp.twilio.com/browse/VBLOCKS-146)

### Description

Fix double register during `Device.updateOptions`.

### Before review
* [x] Ready for review

### Before merge
* [x] Got one or more +1s
* [x] Squashed erroneous commits if necessary
* [x] Re-tested if necessary

### DIFF
--- a/tests/unit/device.ts
+++ b/tests/unit/device.ts
@@ -377,6 +377,30 @@ describe('Device', function() {
           sinon.assert.calledOnce(setupStreamSpy);
         });
 
+        it('should not throw during re-registration', async () => {
+          await registerDevice();
+
+          const regCalls: Array<Promise<any>> = [];
+
+          device.register = () => {
+            const regCall = Device.prototype.register.call(device);
+            regCalls.push(regCall);
+            return regCall;
+          };
+
+          device.updateOptions({ edge: 'sydney' });
+
+          pstream.emit('offline');
+          await clock.tickAsync(0);
+          pstream.emit('connected', { region: 'EU_IRELAND' });
+          await clock.tickAsync(0);
+          pstream.emit('ready');
+          await clock.tickAsync(0);
+
+          assert(regCalls.length);
+          await Promise.all(regCalls);
+        });
+
         describe('log', () => {
           let setDefaultLevelStub: any;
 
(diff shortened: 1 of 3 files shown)


tests/unit/device.ts
```diff
@@ -377,6 +377,30 @@ describe('Device', function() {
...
+          };
+
+          device.updateOptions({ edge: 'sydney' });
+
+          pstream.emit('offline');
+          await clock.tickAsync(0);
+          pstream.emit('connected', { region: 'EU_IRELAND' });
+          await clock.tickAsync(0);
+          pstream.emit('ready');
+          await clock.tickAsync(0);
```
charliesantos, line 398: Can we make these clock ticks synchronous with real values instead of zeroes? That way, it's closer to the real behavior.

charliesantos (changes requested):
We should add integration tests.

tests/unit/device.ts
mhuynh5757: Synchronous ticks won't fire off `async` functions properly, this test won't work with `clock.tick` due to the internals of `Device` using `async` functions. I use `0` because I just want the stack to run all immediate `async` functions.

mhuynh5757:
> We should add integration tests.

`m

mhuynh5757:
> We should add integration tests.

The `mutable options > should update edge` test is the closest integration test. Unfortunately I can't think of a good way to test this without using this sort of override, is that OK in integration tests?

charliesantos:
>The mutable options > should update edge test is the closest integration test. Unfortunately I can't think of a good way to test this without using this sort of override, is that OK in integration tests?

We either refactor this so we can properly test, or do the override. I think it's better than not testing it.

charliesantos (approved):
Have we verified these tests fails without the fix? If so, I'm ok merging this PR.

mhuynh5757:
> Have we verified these tests fails without the fix? If so, I'm ok merging this PR.

Yeah, verified manually that both the unit and int tests work with this patch and fail without.

ryan-rowland (approved):

PR #23, opened 2021-08-09
Add Call#reconnected and Call#reconnecting doc blocks

ryan-rowland:
Adds the missing event documentation per https://issues.corp.twilio.com/browse/VBLOCKS-184

### DIFF
--- a/lib/twilio/call.ts
+++ b/lib/twilio/call.ts
@@ -1296,6 +1296,22 @@ namespace Call {
   declare function muteEvent(isMuted: boolean, call: Call): void;
 
   /**
+   * Emitted when the {@link Call} has regained media connectivity.
+   * @param call - The {@link Call}.
+   * @example `call.on('reconnected', () => { })`
+   * @event
+   */
+  declare function reconnected(): void;
+
+  /**
+   * Emitted when the {@link Call} has lost media connectivity and is reconnecting.
+   * @param call - The {@link Call}.
+   * @example `call.on('reconnecting', error => { })`
+   * @event
+   */
+  declare function reconnectingEvent(error: TwilioError): void;
+
+  /**
    * Emitted when the {@link Call} is rejected.
    * @param call - The {@link Call}.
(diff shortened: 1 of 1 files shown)


lib/twilio/call.ts
```diff
@@ -1296,6 +1296,22 @@ namespace Call {
    */
   declare function muteEvent(isMuted: boolean, call: Call): void;
 
+  /**
+   * Emitted when the {@link Call} has regained media connectivity.
+   * @param call - The {@link Call}.
+   * @example `call.on('reconnected', () => { })`
+   * @event
+   */
+  declare function reconnected(): void;
```
mhuynh5757, line 1305:
```suggestion
  declare function reconnectedEvent(call: Call): void;
```
Does this event need to be post-fixed with `-event`?

lib/twilio/call.ts
```diff
@@ -1296,6 +1296,22 @@ namespace Call {
...
+
+  /**
+   * Emitted when the {@link Call} has lost media connectivity and is reconnecting.
+   * @param call - The {@link Call}.
+   * @example `call.on('reconnecting', error => { })`
+   * @event
+   */
+  declare function reconnectingEvent(error: TwilioError): void;
```
mhuynh5757, line 1313: Should we also emit the call?

lib/twilio/call.ts
```diff
@@ -1296,6 +1296,22 @@ namespace Call {
...
+   * Emitted when the {@link Call} has regained media connectivity.
+   * @param call - The {@link Call}.
+   * @example `call.on('reconnected', () => { })`
+   * @event
+   */
+  declare function reconnectedEvent(call: Call): void;
+
+  /**
+   * Emitted when the {@link Call} has lost media connectivity and is reconnecting.
+   * @param call - The {@link Call}.
```
mhuynh5757, line 1309: Should this be `error` and not `call`?

mhuynh5757 (approved):

PR #35, opened 2021-09-27
Signaling reconnection

ryan-rowland:
Signaling reconnection implementation. Network tests are being worked on separately. They will be merged into this branch via PR before merging this branch. Waiting on production updates to run network tests.

### DIFF
--- a/lib/twilio/call.ts
+++ b/lib/twilio/call.ts
@@ -11,6 +11,7 @@ import {
   GeneralErrors,
   InvalidArgumentError,
   MediaErrors,
+  SignalingErrors,
   TwilioError,
   UserMediaErrors,
 } from './errors';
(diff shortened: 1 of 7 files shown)


lib/twilio/call.ts
```diff
@@ -539,13 +550,17 @@ class Call extends EventEmitter {
...
 
-      const onAnswer = (pc: RTCPeerConnection) => {
+      const onAnswer = (pc: RTCPeerConnection, reconnectToken?: string) => {
         // Report that the call was answered, and directionality
         const eventName = this._direction === Call.CallDirection.Incoming
           ? 'accepted-by-local'
           : 'accepted-by-remote';
         this._publisher.info('connection', eventName, null, this);
 
+        if (typeof reconnectToken ===  'string') {
```
mhuynh5757, line 560: Should we remove this check? I'm thinking of a scenario where the call is established, a reconnect token is given to the SDK, the call is network interrupted, the call reconnects, but for some reason the SDK doesn't receive a new reconnect token. Would we want to leave the old reconnect token in the call? It might get used again improperly.

ryan-rowland: Hmm. I'm trying to think of a scenario where this may happen. This should only be called on the answer, so normal WebSocket reconnections wouldn't trigger this. Did you have any scenarios in mind?

mhuynh5757: Ah, I might not be understanding the VSP flow properly. Do we get a `connect` back when we reconnect a call?

PR #38, opened 2021-10-08
Stop current sounds when updating

r0stig:
**Contributing to Twilio**

### Description

Stops currently playing sound when updating device sounds.

## Burndown

In current version if you get an incoming call and update the sounds before answering the old sound will keep playing. This PR stops the current cached sound.

Note: I tried to play the new sound when updating sounds if the isPlaying property is true, but it seems like this property is true in some cases when sounds isn't playing so I couldn't rely on this property in this case.

### Before review
* ~[ ] Updated CHANGELOG.md if necessary~
* ~[ ] Added unit tests if necessary~
* ~[ ] Updated affected documentation~
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/device.ts
+++ b/lib/twilio/device.ts
@@ -744,13 +744,11 @@ class Device extends EventEmitter {
         maxDuration: soundDef.maxDuration,
         shouldLoop: soundDef.shouldLoop,
       });
-      
-      const currentSound = this._soundcache.get(name as Device.SoundName);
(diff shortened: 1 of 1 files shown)


charliesantos:
We're not able to reproduce the issue described in this PR. If you're still seeing it, please submit a bug instead and we will investigate.

Issue #34, opened 2021-09-13
call.ignore() not working in this new version


riteshsainju:
`call.ignore()
`
This method was supposed to the stop incoming sound but its not working after migrating from  1.X version to 2.x version.

riteshsainju:
Any update for this issue?

skylord123:
Same problem here. `call.ignore()` does not work and seems to do nothing.

skylord123:
As a workaround until this gets fixed (a bit sad there hasn't been any activity on this at all) you can do this:

```
device._soundcache.get(Device.SoundName.Incoming).stop();
```

I added this after my `call.ignore()` to stop the ringing sound manually.

ryan-rowland:
Thanks for bringing this to our attention. It seems some logic was unintentionally removed as part of the transition. I've cut an internal ticket to fix this.

charliesantos:
Should be fixed as of 2.1.1

PR #69, opened 2022-01-26
GH34 | VBLOCKS-474

ryan-rowland:
@mhuynh5757 Addresses the bug filed here: #34

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+2.1.1 (In Progress)
+===================
+
+Bug Fixes
+---------
+
+- Ignoring a call will now properly stop the ringing sound
+- NPM versioning has been fixed to specify >=12 rather than exactly 12
+
 2.1.0 (December 16, 2021) - Release
 ===================================
 
--- a/lib/twilio/call.ts
+++ b/lib/twilio/call.ts
@@ -225,6 +225,11 @@ class Call extends EventEmitter {
   private readonly _monitor: StatsMonitor;
 
   /**
+   * Method to be run after {@link Call.ignore} is called.
+   */
+  private _onIgnore: () => void;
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+2.1.1 (In Progress)
+===================
+
+Bug Fixes
+---------
+
+- Ignoring a call will now properly stop the ringing sound
+- NPM versioning has been fixed to specify >=12 rather than exactly 12
```
mhuynh5757, line 8: Is this change included in this PR?

lib/twilio/call.ts
```diff
@@ -1550,6 +1560,11 @@ namespace Call {
      */
     isUnifiedPlanDefault: boolean;
 
+    /**
+     * A function to be called after {@link Call.ignore} is called.
+     */
+    onIgnore: () => any;
```
mhuynh5757, line 1566: The other type is `() => void` (which I believe is preferred over `any`). Should we update this one to match?

CHANGELOG.md
ryan-rowland: This was added on master and the CHANGELOG wasn't updated, so just adding it here.

lib/twilio/call.ts
```diff
@@ -689,6 +698,7 @@ class Call extends EventEmitter {
     this._status = Call.State.Closed;
     this._mediaHandler.ignore(this.parameters.CallSid);
     this._publisher.info('connection', 'ignored-by-local', null, this);
+    this._onIgnore();
```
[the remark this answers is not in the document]
ryan-rowland: We always pass it, but you're right. We should treat `Call` as an independent class as it is. Will update.

PR #94, opened 2022-06-11
Add status event, corresponding unit tests, and update documentation

henryStelle:
**Contributing to Twilio**

### Description

Adds a status event to the Call object that fires whenever the `_status` private variable is updated. This is important for implementation in React projects so that your UI can be updated when the status changes without requiring manual checks in an interval.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+2.1.2 (???? ??, 2022)
+=========================
+
+New Features
+---------
+
+### Call status event
+The SDK now allows you to subscribe to changes to the `Call` object's state using the `status` event.
+
+```ts
+const call = await device.connect(...);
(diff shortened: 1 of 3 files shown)


henryStelle:
@kbagchiGWC @bobiechen-twilio @makarandp0 @maxterry @kwhinnery @mattmakai would you please review this PR. This would be very helpful functionality.

charliesantos:
Thanks for submitting! The current implementation in the SDK already allows you to monitor status changes. We emit an event related to the status change, instead of a generic 'status' event. Though I would agree that this is a nice to have feature. We are going to consider this in a future release instead.

Issue #95, opened 2022-06-13
[BUG] Twilio error code not being returned when microphone access is blocked


chsweet:
- [x] I have verified that the issue occurs with the latest twilio.js release and is not marked as a known issue in the [CHANGELOG.md](https://github.com/twilio/twilio-voice.js/blob/master/CHANGELOG.md).
 - [x] I reviewed the [Common Issues](https://github.com/twilio/twilio-voice.js/blob/master/COMMON_ISSUES.md) and open GitHub issues and verified that this report represents a potentially new issue.
 - [x] I verified that the [Quickstart](https://github.com/TwilioDevEd/voice-javascript-sdk-quickstart-node) application works in my environment.
 - [x] I am not sharing any [Personally Identifiable Information (PII)](https://www.twilio.com/docs/glossary/what-is-personally-identifiable-information-pii)
       or sensitive account information (API keys, credentials, etc.) when reporting this issue.

**Code to reproduce the issue:**

Block microphone access in browser. Create a new Device

``` js script
  const { Device } = import('@twilio/voice-sdk');
  device = new Twilio.Device(token);

  device.on('error', twilioError => {
    console.log(twilioError);
  });

  device.connect();
```

**Expected behavior:**

When a new `Device` is created and the user has denied access to microphone we should be receiving the Twilio error code 31401 (updated from [31208](https://www.twilio.com/docs/voice/sdks/javascript/migrating-to-js-voice-sdk-20#update-error-codes:~:text=and%20other%20SDKs.-,Update%20error%20codes,-With%20the%20transition)) should be returned.

**Actual behavior:**

We are not receiving the Twilio error code when the user has denied access to microphone. Instead we are getting an error. Slightly different in Chrome, Firefox, and Safari.

**Chrome**
  error  `Uncaught (in promise) TypeError: Cannot ready properties of null (reading 'updatePreferredURI'`

![Screen Shot 2022-06-13 at 12 54 19 PM](https://user-images.githubusercontent.com/81600649/173436957-807ab96f-6d82-497c-bd08-af4625d74f02.png)

**Firefox** 
error `Uncaught (in promise) TypeError: f._stream is null`

<img width="595" alt="Screen Shot 2022-06-13 at 4 37 47 PM" src="https://user-images.githubusercontent.com/81600649/173460061-327c7fd3-7752-463e-aa09-1da2f54b3ada.png">

**Safari** 
[...]

PR #106, opened 2022-08-18
Check if stream exists before unsetting preferred URI

SiimMardus:
**Contributing to Twilio**

### JIRA link(s):

-

### Description

#95

maybeUnsetPreferredUri function was added in this commit:
https://github.com/twilio/twilio-voice.js/commit/b2b65437681ce6aa69870505a40d9f42323d7c9e

This caused unhandled error for some flows, such as when user
denies microphone access.

To fix this, we should check if the stream even exists before
attempting to calling `updatePreferredURI` on it.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [x] Squashed erroneous commits if necessary
* [x] Re-tested if necessary

### DIFF
--- a/lib/twilio/device.ts
+++ b/lib/twilio/device.ts
@@ -977,7 +977,7 @@ class Device extends EventEmitter {
     }, options);
 
     const maybeUnsetPreferredUri = () => {
-      if (this._activeCall === null && this._calls.length === 0) {
+      if (this._stream && this._activeCall === null && this._calls.length === 0) {
         this._stream.updatePreferredURI(null);
       }
     };
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "@types/node": "13.1.8",
     "@types/sinon": "9.0.5",
     "@types/ws": "7.2.0",
+    "@types/mime": "2.0.3",
     "babel-cli": "6.26.0",
     "babel-eslint": "10.0.3",
     "babel-plugin-envify": "1.2.1",
--- a/tests/unit/device.ts
+++ b/tests/unit/device.ts
@@ -786,6 +786,16 @@ describe('Device', function() {
             device.calls[0].emit('error');
             sinon.assert.calledOnceWithExactly(spy, null);
           });
+
+          it('should not unset the preferred uri if stream is null', () => {
+            const spy: any = device['_stream'].updatePreferredURI =
+              sinon.spy(device['_stream'].updatePreferredURI);
+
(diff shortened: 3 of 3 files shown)


SiimMardus:
@mhuynh5757 @charliesantos @ryan-rowland Could anyone review this PR please 🙂 Currently when user denies media permissions, it causes an unhandled error from device, since Device tries to unset preferred URI for Stream when it is `null`.

Simply checking if Stream exists before doing it temporarily fixes the issue in this flow. I lack the knowledge whether or not the stream _should_ exist there at this point, but if it does, it's a different issue that could be looked into after this current fix is in place, as it's blocking us from migrating from V1 to V2.

lib/twilio/device.ts
```diff
@@ -977,7 +977,7 @@ class Device extends EventEmitter {
     }, options);
 
     const maybeUnsetPreferredUri = () => {
-      if (this._activeCall === null && this._calls.length === 0) {
+      if (this._stream && this._activeCall === null && this._calls.length === 0) {
```
charliesantos, line 980: Can we please log a warning if stream is null? Maybe put this check at the top?
```js
    const maybeUnsetPreferredUri = () => {
      if (!this._stream) {
        return this._log.warn('UnsetPreferredUri called without a stream');
      }
      if (this._activeCall === null && this._calls.length === 0) {
        this._stream.updatePreferredURI(null);
      }
    };
```

charliesantos (changes requested):
This is great @SiimMardus . Thank you for contributing. And apologies for responding too late. I left a comment. Please let me know what you think.

package.json
```diff
@@ -69,6 +69,7 @@
     "@types/node": "13.1.8",
     "@types/sinon": "9.0.5",
     "@types/ws": "7.2.0",
+    "@types/mime": "2.0.3",
```
mhuynh5757, line 72: Let's do a separate PR for this since we should also update our `package-lock.json`.

mhuynh5757 (changes requested):

ostap0207:
@mhuynh5757 Siim and I work together, I have opened another PR that only contains the fix, as changes in these PR are already in our master's fork https://github.com/twilio/twilio-voice.js/pull/143.

charliesantos:
Thanks for submitting @SiimMardus  . I resubmitted your PR here https://github.com/twilio/twilio-voice.js/pull/144 and added our original comments and also a changelog. I'll close this now.

PR #111, opened 2022-09-07
VBLOCKS-1035 Fixing build

charliesantos:
**Contributing to Twilio**

## Pull Request Details

Our build process is failing on node 16 due to outdated dependencies. The 2 main big ones are typescript and typedoc. However, updating typedoc to the working version for our project requires changing how we document most of our files, including a breaking change where we must change how we export namespaces and classes. The latest versions of typedoc does not handle declaration merging very well. Details below.
- Update typedoc, update docstrings, do not update exports - works code wise, but breaks documentation. See attached screenshot below.
- Update typedoc, update docstrings, update exports - documentation is fixed, but we now have breaking changes due to to difference in exports

You can see my progress here https://github.com/twilio/twilio-voice.js/compare/VBLOCKS-1035-update-deps

For this PR, I'm fixing the minimum required to get the build and pipelines working, which includes:
- Adding @types/mime 3.0.0. Later versions is breaking the build. The breakage happens because we don't have a lock file which causes peerDependencies to pull in the latest versions.
- Adding lock file to prevent issues mentioned above.
- Disabling docker tests. They are skipped on firefox and broken on chrome. We will address this on a separate ticket.
- Fixing npm/build badges

For future work, we need to bump the major version in order to update all dependencies to latest version.

EDIT: I forgot to mention, the build was failing when removing `legacy-peer-deps` flag. Adding it back the build only fails due to the @types/mime issue I mentioned above. So we still need to keep `legacy-peer-deps`, unfortunately.

Screen Shot 2022-09-06 at 1 48 22 PM


mhuynh5757:
By major version, do you mean `x` in `x.y.z`?

charliesantos:
> By major version, do you mean `x` in `x.y.z`?

Yes. Basically VoiceJS 3.x 😢

.circleci/images/docker-compose.yml
```diff
@@ -30,7 +30,7 @@ services:
       - /opt/app/node_modules
   integrationTests: # runs integration tets. Expects that sources are mounted.
     <<: *runtimeDefaults
-    command: bash -c "npm install --no-optional --no-legacy-peer-deps && npm run build && ls -la /root && ls -la /root/.npm && npm run test:network"
+    command: bash -c "npm install --legacy-peer-deps && npm run build && ls -la /root && ls -la /root/.npm && npm run test:network"
```
mhuynh5757, line 33: This whole file probably needs reworking.

.circleci/images/docker-compose.yml
```diff
@@ -30,7 +30,7 @@ services:
       - /opt/app/node_modules
   integrationTests: # runs integration tets. Expects that sources are mounted.
```
mhuynh5757, line 31:
```suggestion
  integrationTests: # runs integration tests. Expects that sources are mounted.
```

.gitignore
```diff
@@ -2,7 +2,6 @@
 .idea
 .nyc_output/
 config.yaml
-package-lock.json
```
mhuynh5757, line 5: 🎉

mhuynh5757 (approved):

PR #114, opened 2022-09-21
feat: send message api

mhuynh5757:
**Contributing to Twilio**

### JIRA link(s):

- [VBLOCKS-1075](https://issues.corp.twilio.com/browse/VBLOCKS-1075)

### Description

This PR adds the send message API for SDK Eventing.

### Before review
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/call.ts
+++ b/lib/twilio/call.ts
@@ -10,6 +10,7 @@ import DialtonePlayer from './dialtonePlayer';
 import {
   GeneralErrors,
   InvalidArgumentError,
+  InvalidStateError,
   MediaErrors,
   SignalingErrors,
   TwilioError,
@@ -21,6 +22,7 @@ import RTCSample from './rtc/sample';
 import RTCWarning from './rtc/warning';
 import StatsMonitor from './statsMonitor';
 import { isChrome } from './util';
+import { generateVoiceEventSid } from './uuid';
 
 const Backoff = require('backoff');
 const C = require('./constants');
@@ -236,6 +238,7 @@ class Call extends EventEmitter {
     MediaHandler: PeerConnection,
     offerSdp: null,
     shouldPlayDisconnect: () => true,
+    voiceEventSidGenerator: generateVoiceEventSid,
   };
 
   /**
@@ -279,6 +282,11 @@ class Call extends EventEmitter {
   private _status: Call.State = Call.State.Pending;
 
   /**
+   * Voice event SID generator, creates a unique voice event SID.
+   */
+  private _voiceEventSidGenerator: () => string;
+
+  /**
    * Whether the {@link Call} has been connected. Used to determine if we are reconnected.
    */
   private _wasConnected: boolean = false;
@@ -312,6 +320,9 @@ class Call extends EventEmitter {
     if (this._options.reconnectToken) {
       this._signalingReconnectToken = this._options.reconnectToken;
     }
+
+    this._voiceEventSidGenerator =
+      this._options.voiceEventSidGenerator || generateVoiceEventSid;
 
     this._direction = this.parameters.CallSid ? Call.CallDirection.Incoming : Call.CallDirection.Outgoing;
 
@@ -849,6 +860,45 @@ class Call extends EventEmitter {
       const error = new GeneralErrors.ConnectionError('Could not send DTMF: Signaling channel is disconnected');
       this.emit('error', error);
     }
+  }
+
+  /**
+   * Send a message to Twilio.
+   *
+   * @example
+   * ```ts
+   * call.sendMessage(Call.MessageType.UserDefinedMessage, {
+   *   ahoy: 'world!,
+   * });
+   * ```
+   *
+   * @param messageType - The type of the message to send to Twilio.
+   * @param content - The content of the message to send to Twilio.
+   */
+  sendMessage(messageType: Call.MessageType, content: any) {
+    if (typeof messageType !== 'string') {
+      throw new InvalidArgumentError(
+        'Message type must be an enumeration value of `Call.MessageType` or ' +
+        'a string.',
+      );
+    }
+
+    if (this._pstream === null) {
+      throw new InvalidStateError(
+        'Could not send CallMessage; Signaling channel is disconnected',
+      );
+    }
+
+    const callSid = this.parameters.CallSid;
+    if (typeof callSid !== 'string') {
+      throw new InvalidStateError(
+        'Could not send CallMessage; Call has no CallSid',
+      );
+    }
+
+    const voiceEventSid = this._voiceEventSidGenerator();
+
+    this._pstream.sendMessage(callSid, voiceEventSid, messageType, content);
   }
 
   /**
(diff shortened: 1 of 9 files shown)


lib/twilio/call.ts
```diff
@@ -851,6 +862,45 @@ class Call extends EventEmitter {
     }
   }
 
+  /**
+   * Send a message to Twilio.
```
charliesantos, line 866: Need to expand more on this. Just calling it out. No need to address now. We'll do it as part of polishing up the docs.

lib/twilio/call.ts
```diff
@@ -851,6 +862,45 @@ class Call extends EventEmitter {
...
+   * call.sendMessage(Call.MessageType.UserDefinedMessage, {
+   *   ahoy: 'world!,
+   * });
+   * ```
+   *
+   * @param messageType - The type of the message to send to Twilio.
+   * @param content - The content of the message to send to Twilio.
+   */
+  sendMessage(messageType: Call.MessageType, content: any) {
+    if (typeof messageType !== 'string') {
```
charliesantos, line 879: Should we check for empty strings?

lib/twilio/call.ts
```diff
@@ -851,6 +862,45 @@ class Call extends EventEmitter {
...
+   * ```ts
+   * call.sendMessage(Call.MessageType.UserDefinedMessage, {
+   *   ahoy: 'world!,
+   * });
+   * ```
+   *
+   * @param messageType - The type of the message to send to Twilio.
+   * @param content - The content of the message to send to Twilio.
+   */
+  sendMessage(messageType: Call.MessageType, content: any) {
```
charliesantos, line 878: Should this be a promise? Or is that a separate task? Right now it just a straight hand off to signaling.

lib/twilio/call.ts
```diff
@@ -851,6 +862,45 @@ class Call extends EventEmitter {
...
+        'Could not send CallMessage; Signaling channel is disconnected',
+      );
+    }
+
+    const callSid = this.parameters.CallSid;
+    if (typeof callSid !== 'string') {
+      throw new InvalidStateError(
+        'Could not send CallMessage; Call has no CallSid',
+      );
+    }
```
charliesantos, line 897: I think it's better to check call status so you don't need these 2 checks. You will also probably cover more scenario. Something like
```ts
if (this._status !== Call.State.Open) {
  throw new InvalidStateError(
      `Could not send message. Call status is ${this._status}`
  );
}
``

File: lib/twilio/pstream.js [hunk 5]
```diff
@@ -177,6 +177,22 @@ PStream.prototype.setToken = function(token) {
   this._publish('listen', payload);
 };
 
+PStream.prototype.sendMessage = function(
```
charliesantos, line 180: Something to keep at the back of our heads. This pattern is old school and needs rework.

package.json
```diff
@@ -122,8 +122,10 @@
   "dependencies": {
     "@twilio/audioplayer": "1.0.6",
     "@twilio/voice-errors": "1.1.1",
+    "@types/md5": "^2.3.2",
     "backoff": "2.5.0",
     "loglevel": "1.6.7",
+    "md5": "^2.3.0",
```
charliesantos, line 128: Let's use exact version for consistency. I know we now have a lock file, but we should keep this consistent.

charliesantos: What is the size difference?

lib/twilio/uuid.ts
```diff
@@ -0,0 +1,39 @@
+import * as md5 from 'md5';
+import { NotSupportedError } from '../twilio/errors';
+
+export function generateUuid(): string {
```
charliesantos, line 4: Why are we exporting it? For unit tests? If so, we should still keep it private. We can hit this function by testing generateVoiceEventSid instead.

charliesantos (changes requested):

lib/twilio/call.ts
mhuynh5757: Good point, I will add that.

lib/twilio/call.ts
mhuynh5757: Yeah so that's the task that's rolling over, using the `ack` from signaling.

lib/twilio/call.ts
mhuynh5757: Sounds good, will do.

package.json
mhuynh5757: What's the best way to calculate that? Just look at the size of the two new packages?

lib/twilio/uuid.ts
mhuynh5757: Hmm, I can see us using in other things down the line maybe, but we can export it then. I'll make it non-exported.

package.json
charliesantos: Build the package, check the size of the minified/unminified file and compare it to the previous release (download from tags or npm)

lib/twilio/uuid.ts
charliesantos: Right. We should not export anything that is not needed. Only do it when you need it. [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) 😉

lib/twilio/call.ts
mhuynh5757: Do you think it's still worth type-guarding `this.parameters.CallSid`?

charliesantos: Do you mean keep the `typeof callSid !== 'string'` check? I don't think we need it. Check the status that it is open should cover that.

package.json
mhuynh5757: `md5` and `@types/md5` in the end adds 10kb; `twilio.min.js` goes from 335kb to 345kb

lib/twilio/call.ts
charliesantos: @mhuynh5757 I just learned base on this [discussion](https://twilio.slack.com/archives/C02U2QQCMME/p1663964471252479) that client should be able to receive events outside of "open" status, such as ringing. This means, we should only check here if there's a websocket connection or not.

mhuynh5757: What about a CallSid? This is sending not receiving right? Seems kind of weird to want to send a message during ringing.

charliesantos: It's a valid use case. The customer can use it to send status events to their backend or other clients.

mhuynh5757: Ok, sounds solid then.

mhuynh5757: Do we want to add the CallSid check back in?

charliesantos: No, we don't want to check for callsid. I don't think callsid is generated until the call is established no?

mhuynh5757: Yes, but we always send the `CallSid` with every `message` package over VSP.

charliesantos:
Latest changes looks good. Please address the callsid related discussion then we can merge
https://twilio.slack.com/archives/C042FCK9WSD/p1663968312559099?thread_ts=1663961297.212249&cid=C042FCK9WSD

charliesantos (approved):

charliesantos:
I'll go ahead and merge this one @mhuynh5757

Issue #14, opened 2021-07-15
No mechanism for disabling sounds


doryphores:
v1 had an API to enable/disable sounds:

```
Device.audio.incoming(false)
```

I can see some code related to this in v2: https://github.com/twilio/twilio-voice.js/blob/385c31fc7a2492f7b4f279af3816aa36365597b3/lib/twilio/device.ts#L349-L353

But there doesn't seem to be a public API to configure the settings.

Is this an oversight? It is currently blocking us from upgrading.

ryan-rowland:
Hi @doryphores,
Thanks for bringing this to our attention. It looks like when we deprecated the old `Device.setup()` interfaced we missed a step in wiring up the Device to check the AudioHelper's settings rather than its internal (and unchangeable) default settings.

I'm creating an internal ticket for this and you should expect this to be fixed in an upcoming patch release.

cjstewart88:
@ryan-rowland is there an update on this?

cjstewart88:
Any update on this?

doryphores:
@ryan-rowland Any update on this? We really really want to upgrade 😄

kamalbennani:
@ryan-rowland any news on this?

jakswa:
Maybe impacted by the recent twilio layoffs?

kamalbennani:
I'm obliged to make the "_enabledSounds" public to disable the outgoing and incoming sounds:
```js
device._enabledSounds[Device.SoundName.Outgoing] = false;
```
Not great, but at least it does the trick.

kamalbennani:
@charliesantos by checking the [code](https://github.com/twilio/twilio-client.js/blob/master/lib/twilio/audiohelper.ts#L362) I see that everything is already implemented, the only thing missing is to type the Audio class correctly, if that's the case, I can push a PR to fix the typing issue, is that fine with you?

charliesantos:
@kamalbennani we have not looked into it yet but please feel free to submit a PR :)
I believe the fix is to update the references of `_enabledSounds` in the device class to use the values set in the `_audio` object. For example, this line needs needs updating https://github.com/twilio/twilio-voice.js/blob/master/lib/twilio/device.ts#L995 to use something like `this._audio.outgoing()`.

doryphores:
[...]

PR #115, opened 2022-10-03
[DRAFT] move enable sounds to the audio helper class and expose getter and setters to enable and disable toggeable sounds

kamalbennani:
The goal of this PR is to answer the "lack of implementation" described in this [issue](https://github.com/twilio/twilio-voice.js/issues/14)

**Contributing to Twilio**

### JIRA link(s):

### Description

This PR introduces a couple of changes:
- Remove enabledSounds responsibility from the Device class and move it close to the AudioHelper
- Expose toggleable sounds getters/setters
- Use `this.audio.outgoing` instead of relying on the enabledSounds private property

### Before review
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/audiohelper.ts
+++ b/lib/twilio/audiohelper.ts
@@ -21,6 +21,12 @@ const kindAliases: Record<string, string> = {
   audiooutput: 'Audio Output',
 };
 
+export const defaultEnabledSounds: Record<Device.ToggleableSound, boolean> = {
+  [Device.SoundName.Disconnect]: true,
+  [Device.SoundName.Incoming]: true,
+  [Device.SoundName.Outgoing]: true,
+};
+
 /**
  * Provides input and output audio-based functionality in one convenient class.
  * @publicapi
@@ -30,6 +36,23 @@ class AudioHelper extends EventEmitter {
    * The currently set audio constraints set by setAudioConstraints(). Starts as null.
    */
   get audioConstraints(): MediaTrackConstraints | null { return this._audioConstraints; }
+
+  /**
+   * If no value is passed, it returns a boolean indicating whether the disconnect sound
+   * is enabled or disabled otherwise it enables or disables the disconnect event sound.
+   */
+  [Device.SoundName.Disconnect]: EnableSoundFn;
+
+  /**
+   * If no value is passed, it returns a boolean indicating whether the incoming sound
+   * is enabled or disabled otherwise it enables or disables the incoming event sound.
+   */
+  [Device.SoundName.Incoming]: EnableSoundFn;
+  /**
+   * If no value is passed, it returns a boolean indicating whether the outgoing sound
+   * is enabled or disabled otherwise it enables or disables the outgoing event sound.
+   */
+  [Device.SoundName.Outgoing]: EnableSoundFn;
 
   /**
    * A Map of all audio input devices currently available to the browser by their device ID.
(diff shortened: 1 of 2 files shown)


charliesantos:
@mhuynh5757 can you please do a quick pass?

lib/twilio/audiohelper.ts
```diff
@@ -359,8 +387,8 @@ class AudioHelper extends EventEmitter {
       return enabledSounds[key];
     }
 
-    Object.keys(enabledSounds).forEach(key => {
-      (this as any)[key] = setValue.bind(null, key);
+    Object.keys(enabledSounds).forEach((key: Device.ToggleableSound) => {
```
mhuynh5757, line 390: 👍

lib/twilio/audiohelper.ts
```diff
@@ -172,9 +201,7 @@ class AudioHelper extends EventEmitter {
     this.isOutputSelectionSupported = isEnumerationSupported && isSetSinkSupported;
     this.isVolumeSupported = isAudioContextSupported;
 
-    if (options.enabledSounds) {
-      this._addEnabledSounds(options.enabledSounds);
-    }
+    this._addEnabledSounds();
```
mhuynh5757, line 204: It seems like we only call this once at instantiation, and therefore we only add sounds that are enabled at construction time. It's not immediately clear to me that we can enable sounds post-construction, perhaps we need some mechanism for that?

lib/twilio/audiohelper.ts
```diff
@@ -21,6 +21,12 @@ const kindAliases: Record<string, string> = {
   audiooutput: 'Audio Output',
 };
 
+export const defaultEnabledSounds: Record<Device.ToggleableSound, boolean> = {
```
mhuynh5757, line 24: I think we can remove these defaults from `device` if we want the `AudioHelper` class to have more control over the enablement of sounds.

mhuynh5757: Sorry, didn't see the `device` changes, please ignore this comment! 😓

lib/twilio/device.ts
```diff
@@ -972,7 +963,7 @@ class Device extends EventEmitter {
       maxAverageBitrate: this._options.maxAverageBitrate,
       preflight: this._options.preflight,
       rtcConstraints: this._options.rtcConstraints,
-      shouldPlayDisconnect: () => this._enabledSounds.disconnect,
+      shouldPlayDisconnect: () => this._audio ? this._audio.disconnect() : defaultEnabledSounds.disconnect,
```
mhuynh5757, line 966: We should check if these members are defined, i.e. `this._audio.disconnect` might be undefined depending on the state of the `AudioHelper` and we might try to invoke it.

mhuynh5757:
@kamalbennani thanks for contributing to the SDK! All feedback is greatly appreciated. I left some comments on some more specific details.

One recommendation over the current implementation that I'd like to see change is moving away from setting `EnableSoundFn` members on the object and using something less anti-pattern, like indexing an internal map with the `Sound` enumeration we already have. So the API would look something like ```
audioHelper.shouldPlaySound(Device.SoundName.Incoming)
```
or maybe we relegate all "shouldPlay" logic to the `AudioHelper` entirely and in the `Device` we just ask the `AudioHelper` to play it, getting some sort of return value from the `AudioHelper` if necessary.

What do you think?

charliesantos:
@kamalbennani I'm closing this now. See my comment here https://github.com/twilio/twilio-voice.js/issues/14#issuecomment-1489266693

PR #121, opened 2022-10-24
fix: event publisher no longer working after having called `Device.updateOptions`

rmnbrd:
**Contributing to Twilio**

> All third party contributors acknowledge that any contributions they provide will be made under the same open source license that the open source project is provided under.

- [x] I acknowledge that all my contributions will be made under the project's license.

### JIRA link(s):

- [CLIENT-0000](https://issues.corp.twilio.com/browse/CLIENT-0000)

### Description

When `Device.updateOptions` is called, that creates a new Publisher instance.
And this new instance is not working correctly given the `host` property is missing if the `eventgw` option is not defined.

This leads to the "Metrics" tab of the Console to be completely empty, as you can see on that screenshot:
image (1)

This PR introduces a change in that logic. The `host` property would now be defined given we're getting the value from the previous Publisher instance, or falling back to the default.

### Before review
* [ ] Updated CHANGELOG.md if necessary
* [ ] Added unit tests if necessary
* [ ] Updated affected documentation
* [x] Verified locally with `npm test`
* [x] Manually sanity tested running locally
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Squashed erroneous commits if necessary
* [ ] Re-tested if necessary

PR #122, opened 2022-10-24
VBLOCKS-1155 Make sure publisher host is inited if home region exists

charliesantos:
This is an updated fix for https://github.com/twilio/twilio-voice.js/pull/121

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+2.1.2 (In Progress)
+=========================
+
+Bug Fixes
+---------
+
+- Fixed an issue where insights data stops getting published after calling `device.updateOptions`.
+
 2.1.1 (February 18, 2022)
 =========================
 
--- a/lib/twilio/device.ts
+++ b/lib/twilio/device.ts
@@ -1087,7 +1087,9 @@ class Device extends EventEmitter {
     const region = getRegionShortcode(payload.region);
     this._edge = payload.edge || regionToEdge[region as Region] || payload.region;
     this._region = region || payload.region;
+    this._home = payload.home;
     this._publisher?.setHost(createEventGatewayURI(payload.home));
+
     if (payload.token) {
       this._identity = payload.token.identity;
       if (
@@ -1105,7 +1107,6 @@ class Device extends EventEmitter {
         }, timeoutMs);
       }
     }
-    this._home = payload.home;
 
     const preferredURIs = getChunderURIs(
       this._edge as Edge,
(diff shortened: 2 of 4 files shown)


charliesantos:
@kamalbennani @rmnbrd can you please take a look at this one and see if it fixes your need? This is in response to your PR https://github.com/twilio/twilio-voice.js/pull/121

You are correct that the publisher's host is not updated after calling updateOptions. However, we need to reinitialize from the home region value.

charliesantos:
> It seems like we only set the `this._home` value when we receive a "signaling connected" message, what happens if the token is updated such that `this._home` doesn't reflect the home region of the new token?

Good call @mhuynh5757 . This won't be an issue because it is not possible to update the token of an existing device with a different region. The SDK will emit an error in this case. The region and edge parameter needs to match in order for this to work. Meaning, in order to update the home region, you also need to reinitialize the device which will in turn reinitialize signaling and the `_home` property.

Reference: https://github.com/twilio/twilio-voice.js/blob/master/CHANGELOG.md#twilio-regional-support

kamalbennani:
This would work perfectly fine I guess, it's pretty much the same as the solution suggested by @rmnbrd.
I have a nitpicky comment (that you could definitely ignore) but it would be very nice to have the `this._host` assignment next to this._region/this._edge in the `signalingConnected` [method](https://github.com/twilio/twilio-voice.js/blob/19bbeea1370946212b66cd1d6ccb12c0e3033b6c/lib/twilio/device.ts#L1108)
Other than that LGTM.

kamalbennani:
And btw, would it be possible to release it as soon as you could please 🙇

rmnbrd:
Hello @charliesantos, 

Indeed, the changes of this PR seem to resolve the issue @kamalbennani and I described in https://github.com/twilio/twilio-voice.js/pull/121

Could you please let us know if you have any ETA for the release of that fix. 
Because as @kamalbennani mentioned, this problem is affecting us a lot right now, so we would like to know whether we should deploy our hotfix (in which we patched your SDK), or wait for a release from your side.
Thank you in advance.

mhuynh5757 (approved):

charliesantos:
@kamalbennani @rmnbrd thank you for your input. Romain's PR is good, however, it is using default host which will route data to a default region if we publish right before signaling is connected. This is bad from data privacy perspective.

Moving `_home` assignment next to `_edge`/`_region` is a good idea. I will go ahead and update that.

As for ETA, we will proceed with our internal release process. The patch release should be out after 1 day at the earliest if there are no issues found.

PR #129, opened 2022-12-01
Feature/sdk eventing

charliesantos:
Merge feature branch


kbagchiGWC (approved):

lib/twilio/pstream.js
```diff
@@ -177,6 +177,23 @@ PStream.prototype.setToken = function(token) {
   this._publish('listen', payload);
 };
 
+PStream.prototype.sendMessage = function(
+  callsid,
+  content,
+  contenttype = 'application/json',
```
mhuynh5757, line 183: Probably too late to mention this now, but if I recall correctly, optional arguments should always come last in a function signature. Should we note this for the future?

kbagchiGWC: Any plans to change this now @charliesantos @mhuynh5757? Or should we create a task and take care of it in 1.5 release?

charliesantos: This is internal and more of a cosmetic suggestion. Check it's usage in `call.ts`. It is a big deal if this is the public API.

PR #130, opened 2022-12-10
pstream answer listener memory leak

dimasikturbo:
When "answer" callback is added, a new function is created using bind() method. As a results that listener is never removed from the pstream and over the time when multiple calls are created the number of listeners just keeps growing.

**Contributing to Twilio**

- [ ] I acknowledge that all my contributions will be made under the project's license.

### JIRA link(s):

### Description

A description of what this PR does.

### Before review
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/call.ts
+++ b/lib/twilio/call.ts
@@ -628,15 +628,17 @@ class Call extends EventEmitter {
 
       this._pstream.addListener('hangup', this._onHangup);
 
+      this._onAnswerBind = this._onAnswer.bind(this);
+
       if (this._direction === Call.CallDirection.Incoming) {
         this._isAnswered = true;
-        this._pstream.on('answer', this._onAnswer.bind(this));
+        this._pstream.on('answer', this._onAnswerBind);
         this._mediaHandler.answerIncomingCall(this.parameters.CallSid, this._options.offerSdp,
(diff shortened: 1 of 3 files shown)


lib/twilio/call.ts
```diff
@@ -628,15 +628,17 @@ class Call extends EventEmitter {
...
         this._isAnswered = true;
-        this._pstream.on('answer', this._onAnswer.bind(this));
+        this._pstream.on('answer', this._onAnswerBind);
         this._mediaHandler.answerIncomingCall(this.parameters.CallSid, this._options.offerSdp,
           rtcConstraints, rtcConfiguration, onAnswer);
       } else {
         const params = Array.from(this.customParameters.entries()).map(pair =>
          `${encodeURIComponent(pair[0])}=${encodeURIComponent(pair[1])}`).join('&');
-        this._pstream.on('answer', this._onAnswer.bind(this));
+        this._pstream.on('answer', this._onAnswerBind);
```
kamalbennani, line 641: Can't you just turn the `_onAnswer` to an arrow function (same as the other class methods) to avoid this issue from happening?

kamalbennani (changes requested):

PR #138, opened 2023-01-23
Feature/voice js dns

charliesantos:
Merge feature


kbagchiGWC (commented):
Wow more delete than add :)

lib/twilio/call.ts
```diff
@@ -1860,11 +1855,6 @@ namespace Call {
      */
     reconnectToken?: string;
 
-    /**
-     * The Region currently connected to.
-     */
-    region?: string;
```
mhuynh5757, line 1866: Just making sure, we're OK with this API change for a minor version bump?

mhuynh5757 (approved):
Looks good to me, I don't immediately see any API changes, did I miss anything?

mhuynh5757:
> Wow more delete than add :)

Mostly deprecated things :P I remember adding that years ago, going to miss it 😢 (jk)

charliesantos:
> Looks good to me, I don't immediately see any API changes, did I miss anything?

Yep. No public API change. But if you look at region.ts, we are now directly using the new dns names. Before, we are mapping those to the old uris.

PR #152, opened 2023-03-30
VBLOCKS-1534 | Stop using deprecated RTCIceCandidateStats

charliesantos:
**Contributing to Twilio**

> All third party contributors acknowledge that any contributions they provide will be made under the same open source license that the open source project is provided under.

- [x] I acknowledge that all my contributions will be made under the project's license.

[VBLOCKS-1534](https://issues.corp.twilio.com/browse/VBLOCKS-1534)

### Before review
* [x] Updated CHANGELOG.md if necessary
* [x] Added unit tests if necessary
* [x] Updated affected documentation
* [x] Verified locally with `npm test`
* [x] Manually sanity tested running locally
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Squashed erroneous commits if necessary
* [ ] Re-tested if necessary

PR #153, opened 2023-03-30
VBLOCKS-1111 | Fix sound definitions

charliesantos:
**Contributing to Twilio**

## Pull Request Details

@mhuynh5757 please review and get [this PR ](https://github.com/twilio/twilio-voice.js/pull/152) merged in first.

https://issues.corp.twilio.com/browse/VBLOCKS-1111
Github https://github.com/twilio/twilio-voice.js/issues/14

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/audiohelper.ts
+++ b/lib/twilio/audiohelper.ts
@@ -293,30 +293,33 @@ class AudioHelper extends EventEmitter {
   }
 
   /**
-   * Set whether the disconnect sound is enabled or not
-   * @param isEnabled - Whether the disconnect sound is enabled or not
-   * @returns Whether the disconnect sound is enabled or not
+   * Enable or disable the disconnect sound.
+   * @param doEnable Passing `true` will enable the sound and `false` will disable the sound.
+   * Not passing this parameter will not alter the enable-status of the sound.
+   * @returns The enable-status of the sound.
    */
-  disconnect(isEnabled?: boolean): boolean {
-    return this._maybeEnableSound(Device.SoundName.Disconnect, isEnabled);
+  disconnect(doEnable?: boolean): boolean {
+    return this._maybeEnableSound(Device.SoundName.Disconnect, doEnable);
   }
 
   /**
-   * Set whether the incoming sound is enabled or not
-   * @param isEnabled - Whether the incoming sound is enabled or not
-   * @returns Whether the incoming sound is enabled or not
+   * Enable or disable the incoming sound.
+   * @param doEnable Passing `true` will enable the sound and `false` will disable the sound.
+   * Not passing this parameter will not alter the enable-status of the sound.
+   * @returns The enable-status of the sound.
    */
-  incoming(isEnabled?: boolean): boolean {
-    return this._maybeEnableSound(Device.SoundName.Incoming, isEnabled);
+  incoming(doEnable?: boolean): boolean {
+    return this._maybeEnableSound(Device.SoundName.Incoming, doEnable);
   }
 
   /**
-   * Set whether the outgoing sound is enabled or not
-   * @param isEnabled - Whether the outgoing sound is enabled or not
-   * @returns Whether the outgoing sound is enabled or not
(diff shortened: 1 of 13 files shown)


lib/twilio/audiohelper.ts
```diff
@@ -287,6 +292,33 @@ class AudioHelper extends EventEmitter {
     }
   }
 
+  /**
+   * Set whether the disconnect sound is enabled or not
+   * @param isEnabled - Whether the disconnect sound is enabled or not
+   * @returns Whether the disconnect sound is enabled or not
+   */
+  disconnect(isEnabled?: boolean): boolean {
```
mhuynh5757, line 300: If we pass `undefined` it seems like these functions just report the enable status of the sound, is that correct? Should we put that behavior in the docstring?

lib/twilio/device.ts
```diff
@@ -978,7 +969,7 @@ class Device extends EventEmitter {
       maxAverageBitrate: this._options.maxAverageBitrate,
       preflight: this._options.preflight,
       rtcConstraints: this._options.rtcConstraints,
-      shouldPlayDisconnect: () => this._enabledSounds.disconnect,
+      shouldPlayDisconnect: () => this.audio?.disconnect(),
```
mhuynh5757, line 972: When would `this.audio` be `undefined` or `null`?

lib/twilio/device.ts
```diff
@@ -1213,7 +1204,7 @@ class Device extends EventEmitter {
       this._publishNetworkChange();
     });
 
-    const play = (this._enabledSounds.incoming && !wasBusy)
+    const play = (this.audio?.incoming() && !wasBusy)
       ? () => this._soundcache.get(Device.SoundName.Incoming).play()
```
mhuynh5757, line 1208: It seems like this kind of logic could be placed in the `AudioHelper` class itself. Probably out of scope for this ticket though.

lib/twilio/audiohelper.ts
charliesantos: Yes that's correct. If you looks at the generated API doc, the return value kind of explains that right? That is, it's always returning the enable status of the sound whether you pass a boolean or undefined.

lib/twilio/device.ts
charliesantos: It's initial value is `null`. Or when it gets destroyed and you still have lingering call objects with listeners that references this.audio. Very edgy but we want to fail silently instead of throwing benign errors.

lib/twilio/audiohelper.ts
mhuynh5757: My confusion is that I wasn't sure if passing `undefined` would be inferred as `false` since it's a falsy value, or if it would make it a no-op.

charliesantos: It won't be. See `_maybeEnableSound` implementation in this file. I also added tests for this. Please see the tests.

charliesantos: Ah did you mean something like `disconnect(undefined)` instead of `disconnect()`?

mhuynh5757: It's clear when reading the implementation but not when reading only the docstring, is what I'm trying to say. Applies to either implicit or explicit `undefined`.

charliesantos: Ok gotcha. Can you provide a suggestion on the docstring? What would make it clear?

charliesantos: This is too verbose. 
Re: @param doEnable. Typedoc already generates explanation that this is an optional boolean value. I don't think mentioning it again in the docstring is good.
Re: @returns. Typedoc also mentions that the return value is boolean. No need to mention again.
I tweaked it a bit. Please see latest.

mhuynh5757 (approved):

PR #164, opened 2023-05-09
Merge feature branch for VDI support

charliesantos:
Merge feature branch for VDI support. All changes has already been reviewed, except for changelog.


kbagchiGWC:
The build checks failed. Is that concerning?

.circleci/config.yml
```diff
@@ -78,12 +78,12 @@ commands:
       - store_artifacts:
           path: ./dist
           prefix: ./dist
-  unit-tests:
+  build-checks:
```
mhuynh5757, line 81: Should we split these up in a future ticket?

mhuynh5757 (approved):

.circleci/config.yml
charliesantos: I think it's good to combine the build checks (currently include unit/lint/typechecks)
I don't see benefits splitting them up in CircleCI, only downsides - takes more time, duplicates (in this case 3x) all other pre-install routines such as npm install, loading/saving CI cache, etc which takes too much CI minutes.

PR #168, opened 2023-06-06
VBLOCKS-1578 | Adding typechecks to all public APIs

charliesantos:
**Contributing to Twilio**

### Description

This PR adds typechecks to all public facing APIs

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


tests/typecheck/index.ts
mhuynh5757, line 1: Is this file necessary? Perhaps I am misunderstanding, but the other `tests/typecheck/*` files are simply public API checks, so if we accidentally change our public API then these the Typescript compiler will fail to transpile these files. We don't actually intend to instantiate devices with `"foo"` tokens right?

If these files are also just to ensure that we don't change public API on accident, how do we feel about API rollups similarly to how we do it in the RN SDK?

charliesantos: You are correct, the file is not needed. I just removed it. And you are correct on the other assumptions.

Regarding API rollups, can you point me to the exact location in RN SDK where I can look at it?
For context, this typecheck solution is something that's already there and has also been utilized by video. I want to prevent any scope creep. But I'm interested how it's done in RN sdk

mhuynh5757 (approved):

PR #170, opened 2023-06-08
VBLOCKS-1301 Remove ws and xmlhttprequest dependencies.

manjeshbhargav:
**Contributing to Twilio**

### JIRA link(s):

- [VBLOCKS-1301](https://issues.corp.twilio.com/browse/VBLOCKS-1301)

### Description

* Remove `xmlhttprequest` package from dependency (use `fetch` instead)
* Move `ws` package to `devDependencies` (as it is required for unit tests)

### Confirmation from reviewer (@charliesantos or @mhuynh5757)

* [x] Switching to `fetch` did not disrupt the working of insights reporting

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/request.ts
+++ b/lib/twilio/request.ts
@@ -7,14 +7,13 @@
 
 function request(method, params, callback) {
   const body = JSON.stringify(params.body || {});
-  const fetchUrl = options.fetch || fetch;
   const headers = new Headers();
 
   params.headers = params.headers || [];
   Object.entries(params.headers).forEach(([headerName, headerBody]) =>
     headers.append(headerName, headerBody));
 
-  fetchUrl(params.url, { body, headers, method })
+  fetch(params.url, { body, headers, method })
     .then(response => response.text(), callback)
     .then(responseText => callback(null, responseText), callback);
 }
--- a/lib/twilio/request.ts
+++ b/lib/twilio/request.ts
@@ -4,30 +4,19 @@
  * @internalapi
  */
 // @ts-nocheck
-import { XMLHttpRequest as XHR } from 'xmlhttprequest';
 
 function request(method, params, callback) {
-  const options = {};
-  options.XMLHttpRequest = options.XMLHttpRequest || XHR;
-  const xhr = new options.XMLHttpRequest();
+  const body = JSON.stringify(params.body || {});
+  const fetchUrl = options.fetch || fetch;
+  const headers = new Headers();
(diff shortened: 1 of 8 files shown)


charliesantos (changes requested):
I did a quick pass, here are my initial comments. I'll continue to review later.
- `ws` is used in [wstransport.ts](https://github.com/twilio/twilio-voice.js/blob/VBLOCKS-1301/remove-xmlhttprequest-ws/lib/twilio/wstransport.ts#L8). We should update this file so it uses the browser's default WebSocket class and verify it works on all supported browsers.
- build is failing due to the flaky test. please re-run the job

lib/twilio/request.ts
```diff
@@ -4,30 +4,19 @@
...
  */
 // @ts-nocheck
-import { XMLHttpRequest as XHR } from 'xmlhttprequest';
 
 function request(method, params, callback) {
-  const options = {};
-  options.XMLHttpRequest = options.XMLHttpRequest || XHR;
-  const xhr = new options.XMLHttpRequest();
+  const body = JSON.stringify(params.body || {});
+  const fetchUrl = options.fetch || fetch;
```
charliesantos, line 10:
```suggestion
```

lib/twilio/request.ts
```diff
@@ -4,30 +4,19 @@
...
-
-    callback(new Error(xhr.responseText));
-  };
-  // tslint:disable-next-line
-  for (const headerName in params.headers) {
-    xhr.setRequestHeader(headerName, params.headers[headerName]);
-  }
-
-  xhr.send(JSON.stringify(params.body));
+  fetchUrl(params.url, { body, headers, method })
```
charliesantos, line 17:
```suggestion
  fetch(params.url, { body, headers, method })
```

tests/index.ts
mhuynh5757, line 1: It doesn't look like we use this file for network (e2e) tests, so we don't really need a real `WebSocket` implementation right? Should we think about just mocking the `WebSocket` object so we can drop the dependency entirely?

manjeshbhargav: @mhuynh5757 We already mock the WebSocket for unit tests. Unfortunately the typescript compiler for mocha complains about not only the type definition, but also about the package itself. So, this is the best option as of now.

PR #177, opened 2023-06-27
VBLOCKS-1299 | Making all integration tests green

charliesantos:
**Contributing to Twilio**

## Pull Request Details

See VBLOCKS-1299

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


tests/integration/device.ts
mhuynh5757, line 1: Did these tests just move in the file, or did they change in some way?

mhuynh5757 (approved):

tests/integration/device.ts
charliesantos: Moved and changed.
Moved because if tokenWillExpire test fails, we want it to fail sooner than waiting for all device tests to complete.

For the changes, I changed this section. See the note and the timeout.
```
it('should not emit a "tokenWillExpire" event early', async () => {
      const device = setupDevice(10, 1000);
      await new Promise(async (resolve, reject) => {
        let successTimeout: any = null;
        device.on(Device.EventName.TokenWillExpire, () => {
          if (successTimeout) {
            clearTimeout(successTimeout);
          }
          reject();
        });
        await device.register();
        // NOTE(csantos): Expected time before tokenWillExpire event is raised is about 9s
        // for a ttl: 10s and tokenRefreshMS: 1000, meaning we should resolve the test at less than 9s.
        // However, the ttl returned from signaling, which we use to calculate the timeout,
        // sometimes returns 1s less from the original ttl we provided. So let's account for that.
        // Instead of resolving the test before 9s, we should do it before 8s instead.
        successTimeout = setTimeout(resolve, 7500);
      });
    });
```

PR #182, opened 2023-07-06
VBLOCKS-1301 Add esm output.

manjeshbhargav:
**Contributing to Twilio**

### JIRA link(s):

- [VBLOCKS-1301](https://issues.corp.twilio.com/browse/VBLOCKS-1301)

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
-2.6.1 (July 7, 2023)
+2.7.0 (In Progress)
 ===================
 
+ECMAScript Module Support
+-------------------------
+
+Starting from this version, the SDK can now be imported as an [ECMAScript Module](https://nodejs.org/api/esm.html) (ESM) using the `import` keyword. The SDK can continue to be imported as a [CommonJS Module](https://nodejs.org/api/modules.html) using the `require` keyword.
```
charliesantos, line 7: This might be a little confusing since customers are already able to use `import` with the latest version. Should we make this a bit simpler since we are already providing a link regarding esm? Something like
```
Starting from this version, the SDK can now be imported as an [ECMAScript Module](https://nodejs.org/api/esm.html) (ESM)
```

package.json
```diff
@@ -15,6 +16,10 @@
   "engines": {
     "node": ">= 12"
   },
+  "exports": {
+    "import": "./mjs/twilio.js",
+    "require": "./es5/twilio.js"
```
charliesantos, line 21: Which one is picked up in a typescript project?

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "2.6.2-dev",
   "description": "Twilio's JavaScript Voice SDK",
   "main": "./es5/twilio.js",
+  "module": "./mjs/twilio.js",
   "types": "./es5/twilio.d.ts",
```
charliesantos, line 7: When importing as an esm, does the application imports from es5 folder since the `type` as defined here in package.json points to es5 folder?

package.json
charliesantos: The `main` property (package.json line 5) points to es5. How does it work in conjunction with this section?

package.json
```diff
@@ -26,9 +31,9 @@
     "build:dist": "node ./scripts/build.js ./lib/browser.js ./LICENSE.md ./dist/twilio.js",
     "build:dist-min": "uglifyjs ./dist/twilio.js -o ./dist/twilio.min.js --comments \"/^! twilio-voice.js/\" -b beautify=false,ascii_only=true",
     "build:release": "npm-run-all lint build test:es5 status",
-    "build:ts": "./node_modules/typescript/bin/tsc",
+    "build:ts": "./node_modules/typescript/bin/tsc && ./node_modules/typescript/bin/tsc --project tsconfig-mjs.json",
```
charliesantos, line 34: Since we're now producing mjs folder
- we need to update `.release.json` to also include the folder in the built artifacts.
- should we have a `test:mjs` npm script? See `test:es5`. It validates the es5 files. If it's possible and if you agree, we should also call `test:mjs` wherever we call `test:es5` in the package.json and circileci's config.yml.

charliesantos (changes requested):

package.json
manjeshbhargav: ESM aware bundlers will pick the `./mjs` folder. W.R.T typescript, I think it will pick up from `./mjs` folder if the target specified in `tsconfig.json` is `module`. We need to confirm this though.

manjeshbhargav: > The `main` property (package.json line 5) points to es5. How does it work in conjunction with this section?

`main` is overridden by the above lines in ESM-aware bundlers.

package.json
manjeshbhargav: I could not find a way to specify different directories for types for ESM and CommonJS. Please let me know if you find a way to do it.

package.json
```diff
@@ -43,6 +48,7 @@
     "test": "npm-run-all lint build test:unit test:es5 test:docker",
     "test:docker": "chmod +x ./scripts/run-docker-tests.sh && ./scripts/run-docker-tests.sh",
     "test:es5": "es-check es5 \"./es5/**/*.js\" ./dist/*.js",
+    "test:mjs": "es-check es6 \"./mjs/**/*.js\" --module",
```
charliesantos, line 51: We need to add `test:mjs` on line 33 as well, under `build:release` script.

charliesantos (changes requested):
Left one more comment @manjeshbhargav . Also, the build fails when using svelte+vite. It works with angular and react. The difference I can think of is that, both angular and react uses webpack. Svelte uses vite. You can use my test app here https://github.com/twilio/twilio-voicejs-app/tree/master/frameworks

To properly install the module, in case you haven't already done so, we need to do `npm pack` on voice.js instead of installing from the folder.

1. in twilio-voice.js, run `npm build`, then `npm pack`
2. under any test app I mentioned above e.g. svelte, uninstall voice. `npm uninstall @twilio/voice-sdk`
3. install the package `npm install /path/to/repo/twilio-voice.js/twilio-voice-sdk-2.6.2-dev.tgz`
4. run `npm run dev` to start the server.
5. access the svelete app from your browser, you'll see the error

CHANGELOG.md
```diff
@@ -1,6 +1,25 @@
...
+```ts
+import * as TwilioVoice from '@twilio/voice-sdk/esm';
+```
+
+As of now, ESM support works with the following frameworks:
+
+- React + Webpack
+- Angular + Webpack
+
+If you are using some other framework, you can continue to import the SDK as a [CommonJS Module (CJS)](https://nodejs.org/api/modules.html) using the path `@twilio/voice-sdk`.
```
charliesantos, line 18: ECMAScript Module Support
-----------------------------------
Currently, the SDK is imported as a [CommonJS Module (CJS)](https://nodejs.org/api/modules.html) using the root path `@twilio/voice-sdk`. With this release, the SDK contains an experimental feature that allows it to be imported as an [ECMAScript Module (ESM)](https://nodejs.org/api/esm.html) using the `@twilio/voice-sdk/esm` path. As this is an experimental feature, certain frameworks may not work such as [svelte](https://kit.svelte.dev/docs/introduction). ESM full support will be available in a future release and will become the default import behavior of the SDK.

Example:
```ts
import { Device } from '@twilio/voice-sdk/esm';
```

charliesantos (changes requested):

charliesantos:
@manjeshbhargav esm path does not work @manjeshbhargav . Tested on react and angular https://github.com/twilio/twilio-voicejs-app/tree/master/frameworks

Screen Shot 2023-07-21 at 10 39 17 AM

charliesantos (approved):

PR #192, opened 2023-08-25
[VBLOCKS-2054] Missing generated errors

mhuynh5757:
**Contributing to Twilio**

### JIRA link(s):

- [VBLOCKS-2054](https://issues.corp.twilio.com/browse/VBLOCKS-2054)

### Description

This PR revises the generated error classes and `getErrorsByCode` mapping.

### Screenshots
Screenshot 2023-09-08 at 1 33 00 PM
Screenshot 2023-09-08 at 1 35 47 PM

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
...
+
+- Enhanced error reporting from backend services. Signaling errors emitted by `Device` and `Call` objects should now be more descriptive. If your application logic relied on parsing generic error code `31005`, please consider reading through our API documentation and adjusting your application logic to anticipate a wider and more descriptive range of error codes.
+  - Some new error codes included within this release are:
+    - **ClientErrors**
+      - NotFound: `31404`
+      - TemporarilyUnavailable: `31480`
+      - BusyHere: `31486`
+    - **SIPServerErrors**
+      - Decline: `31603`
+
```
charliesantos, line 17: - Added new errors emitted by `Device` and `Call` objects. Previously, these errors are all part of a generic error code `31005`. With this new release, the following errors now have their own error codes. Please see this [page](https://www.twilio.com/docs/api/errors) for more details about each error.
  - **ClientErrors**
    - NotFound: `31404`
    - TemporarilyUnavailable: `31480`
    - BusyHere: `31486`
  - **SIPServerErrors**
    - Decline: `31603`

_IMPORTANT: If your application relies on listening to the generic error code 31005 when any of the errors above happen, you need to update your error listeners to also listen for the new error codes._

lib/twilio/call.ts
```diff
@@ -1204,7 +1206,10 @@ class Call extends EventEmitter {
 
     this._log.info('Received HANGUP from gateway');
     if (payload.error) {
-      const error = new GeneralErrors.ConnectionError('Error sent from gateway in HANGUP');
+      const code = payload.error.code;
+      const error = typeof code === 'number' && hasErrorByCode(code)
+        ? new (getErrorByCode(code))(payload.error.message)
+        : new GeneralErrors.ConnectionError('Error sent from gateway in HANGUP');
```
charliesantos, line 1212: Do we still publish hangup errors via the device object? If so, should we also update device? If not, can we update the changelog to not mention device?

CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
...
+-------
+
+- Enhanced error reporting from backend services. Signaling errors emitted by `Device` and `Call` objects should now be more descriptive. If your application logic relied on parsing generic error code `31005`, please consider reading through our API documentation and adjusting your application logic to anticipate a wider and more descriptive range of error codes.
+  - Some new error codes included within this release are:
+    - **ClientErrors**
+      - NotFound: `31404`
+      - TemporarilyUnavailable: `31480`
+      - BusyHere: `31486`
+    - **SIPServerErrors**
+      - Decline: `31603`
```
charliesantos, line 16: Are all these new errors surfaced via both Call and Device objects?

charliesantos (changes requested):
Looks good in general @mhuynh5757 . Left a few comments. Also, can you please test the error codes to make sure our parser actually works? Please listen to errors via device and call objects, log the errors, and post a screenshot?

CHANGELOG.md
mhuynh5757: From my testing, no. Just the Call object since the Call object listens to hangup events, not error events. And the new error codes are associated with hangups and as such are only included in hangup type payloads.

lib/twilio/call.ts
mhuynh5757: We don't publish to insights from the device object at all, it seems. Unless I'm missing something. And the Call object publishes all errors it emits to insights, and the hangup errors are emitted and therefore published.

CHANGELOG.md
charliesantos: In that case, we should update the changelog and don't mention `Device`, since these new errors are only emitted by the `Call` object.

lib/twilio/call.ts
charliesantos: Sorry but the use of 'publish' might've caused confusion. I meant 'emit'.
Do we still emit hangup errors via the device object? If so, should we also update device? If not, can we update the changelog to not mention device?

CHANGELOG.md
mhuynh5757: So the thing is, I haven't been able to test `31404` and `31603`, so I'm not sure if the backend sends those as `errors` or as `hangups`. In the first case, if the pstream message type is `error`, then the device would show it, but if it's a `hangup`-associated error, then only the call will show it. I took a look at the voice signaling protocol repo and it doesn't say which error codes can show up in which types of payloads.

lib/twilio/call.ts
mhuynh5757: We do not catch and then re-emit hangup errors with the device, just the call. The device will catch and re-emit payload messages with type `error`, however. See my other comment for my concerns.

lib/twilio/errors/index.ts
```diff
@@ -8,12 +8,51 @@ import {
...
 
+/**
+ * NOTE(mhuynh): Replacing generic error codes with new (more specific) codes,
+ * is a breaking change. If an error code is found in this set, we only perform
+ * the transformation if the feature flag is enabled.
+ *
+ * With every major version bump, such that we are allowed to introduce breaking
+ * changes as per semver specification, this array should be cleared.
+ */
+const FEATURE_FLAG_ERROR_CODES: Set<number> = new Set([
```
mhuynh5757, line 28: @charliesantos should we add _all_ the new error codes to this set?

charliesantos (changes requested):
@mhuynh5757 please address the following then re-request for review.
- finished the items mentioned in our slack thread
- remove this PR from Draft
- update your branch from master

lib/twilio/device.ts
```diff
@@ -1717,6 +1733,71 @@ namespace Device {
      */
     edge?: string[] | string;
 
+    /**
```
mhuynh5757, line 1736: I manually generated and confirmed the autogenerated API docs for this option look good. Perhaps too verbose, wdyt @charliesantos?

charliesantos (changes requested):

lib/twilio/errors/index.ts
```diff
@@ -8,12 +8,81 @@ import {
...
```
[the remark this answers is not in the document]
mhuynh5757: Very good point, I'll make the change.

tests/unit/error.ts
```diff
@@ -49,21 +48,59 @@ describe('Errors', function() {
...
+  it('generated error map matches subset of errors in scripts', () => {
+    const USED_ERRORS: string = require('../../scripts/errors').USED_ERRORS;
+    for (const errorFullName of USED_ERRORS) {
+      const [namespace, name] = errorFullName.split('.');
+      const errorConstructor = (errors as any)[namespace][name];
+      assert(typeof errorConstructor === 'function');
+    }
+  });
+
+  describe('getErrorByFeatureFlagAndCode', () => {
```
charliesantos, line 73: @mhuynh5757 you missed this one

charliesantos (changes requested):

charliesantos (approved):

Issue #197, opened 2023-09-13
[BUG] Crackling audio issue on chrome when zoom is running


charliesantos:
Crackling audio issue on chrome when zoom is running.

Steps to reproduce:
1) Create a zoom meeting and join
2) Make a Twilio Voice Call using the JS SDK
3) Notice the audio crackling sound

SDK version: 2.7.1

We have identified that this is due to a chrome issue https://bugs.chromium.org/p/chromium/issues/detail?id=1474865
To mitigate the issue, we are currently exploring solutions that will prevent calling MediaDevice APIs during a call.

charliesantos:
@kamalbennani-aircall I see your team filed the chromium issue. Is what I described above similar to what you're seeing?

tigranaircall:
Hi @charliesantos. Yes, you're correct, and thank you for pointing that out. The issue you described is indeed similar to what we're experiencing. Our team has filed a Chromium issue to address this problem.
Very good news. We would appreciate it a lot if your team could find a solution.

gruckion:
I also found this is the case too. With the old or new API if I had youtube music open it would cause the Twilio audio to crack.

charliesantos:
fixed in 2.7.2

PR #199, opened 2023-09-19
VBLOCKS-2214 | Remove unnecessary enumerateDevices calls

charliesantos:
**Contributing to Twilio**

## Pull Request Details

https://issues.corp.twilio.com/browse/VBLOCKS-2214
Removing unnecessary enumerateDevices calls every 0.5s. This is to fix a chrome [issue](https://github.com/twilio/twilio-voice.js/issues/197) where audio is choppy when another application is using the audio devices.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


lib/twilio/shims/mediadevices.ts
mhuynh5757, line 1: Love to see us not needing these shims anymore.

lib/twilio/device.ts
```diff
@@ -939,7 +939,15 @@ class Device extends EventEmitter {
...
-      getUserMedia: this._options.getUserMedia || getUserMedia,
+      getUserMedia: (...args) =>
+        (this._options.getUserMedia || getUserMedia)(...args)
+        .then((gumResponse: any) => {
+          this._audio?._updateAvailableDevices().catch(error => {
+            // Ignore error, we don't want to break the call flow
+            this._log.warn('Unable to updateAvailableDevices after gUM call', error);
+          });
+          return Promise.resolve(gumResponse);
+        }),
```
mhuynh5757, line 950: It seems like we're attaching extra responsibility to `gUM` here that it normally is not responsible for. Seems kinda strange, my gut instinct would be to pass the `_audio` to the Call so the Call can perform that logic itself after using `gUM`. Otherwise, this logic is kind of obscure and the Call `gUM` is intrinsically different than other places we use `gUM`.

tests/unit/device.ts
```diff
@@ -78,6 +82,7 @@ describe('Device', function() {
   beforeEach(() => {
     pstream = null;
     publisher = null;
+    updateAvailableDevicesStub = sinon.stub().returns(Promise.reject());
```
mhuynh5757, line 85: If I recall correctly, I had issues when trying something similar to this because the `Promise.reject()` occurs in the closure of the `beforeEach` and not when the stub actually fires. But I could be misremembering...

mhuynh5757 (changes requested):

lib/twilio/audiohelper.ts
```diff
@@ -308,10 +307,41 @@ class AudioHelper extends EventEmitter {
 
     if (this._mediaDevices.removeEventListener) {
       this._mediaDevices.removeEventListener('devicechange', this._updateAvailableDevices);
-      this._mediaDevices.removeEventListener('deviceinfochange', this._updateAvailableDevices);
```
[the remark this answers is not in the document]
charliesantos: Yes.

lib/twilio/audiohelper.ts
```diff
@@ -308,10 +307,41 @@ class AudioHelper extends EventEmitter {
...
     }
   }
 
+  /**
+   * Update the available input and output devices
+   * @private
+   */
+  _updateAvailableDevices = (): Promise<void> => {
+    if (!this._mediaDevices || !this._enumerateDevices) {
+      return Promise.reject('Enumeration not supported');
```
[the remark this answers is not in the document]
charliesantos: This was the original implementation and did not see any reason to change it. The only change that was required is to make it an internal-only public API.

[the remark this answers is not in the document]
mhuynh5757: Ah, I see. Let's leave it as-is then.

lib/twilio/device.ts
charliesantos: I did consider it. However, the Call object does not know anything about the audio helper. The Device object has that responsibility. Putting audio helper inside the Call object means, we're now sharing that between Call and Device, which is not necessary.

We're not attaching extra responsibility to gUM. The Device is responsible for audio devices, including the sequence of MediaDevice's API calls (gUM, enumerateDevices). This implementation is doing that.

Also, no need to call _updateAvailableDevices in other places where we call gUM. We only need it at the start of the call.

tests/unit/device.ts
charliesantos: I did some test to verify this behavior. The rejection does happen when the stub fires. I know, confusing :)

lib/twilio/device.ts
mhuynh5757: Could we rename it from `gUM` to something else? I know it looks cumbersome but I don't like having unknown side-effects.

charliesantos: I think that might create more confusion. Instead, I changed the implementation to use a callback instead. Please take another look.

lib/twilio/call.ts
```diff
@@ -669,6 +669,10 @@ class Call extends EventEmitter {
         data: { audioConstraints },
       }, this);
 
+      if (this._options.onGetUserMedia) {
```
mhuynh5757, line 672: What do you think about `onGetUserMediaResolved`?

mhuynh5757 (approved):

lib/twilio/call.ts
charliesantos: I think it's too verbose..
if you look at existing events, we don't really use past tense. For example, after .accept() is invoked, we emit accept event, no event for failures. Same for other things. Using the convention onEventName has an implicit meaning that an event happened and succeeded.

charliesantos: Just capturing our slack discussion here for future reference. This is more of a JS convention. For example, onClick, and not onClicked. onMouseMove, onTouchStart etc. There are exceptions of course. If we need to capture different stages, then we need to change. For example, connecting, connected, reconnecting, reconnected.

PR #202, opened 2023-09-27
VBLOCKS-2030 | API stubs and error checking

charliesantos:
**Contributing to Twilio**

## Pull Request Details

https://issues.corp.twilio.com/browse/VBLOCKS-2030

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/audiohelper.ts
+++ b/lib/twilio/audiohelper.ts
@@ -3,6 +3,7 @@
  * @module Voice
  */
 import { EventEmitter } from 'events';
+import AudioProcessor from './audioprocessor';
 import Device from './device';
 import { InvalidArgumentError, NotSupportedError } from './errors';
 import Log from './log';
@@ -146,6 +147,11 @@ class AudioHelper extends EventEmitter {
    * Called with the new input stream when the active input is changed.
    */
   private _onActiveInputChanged: (stream: MediaStream | null) => Promise<void>;
+
+  /**
+   * Internal reference to the added AudioProcessor
+   */
+  private _processor: AudioProcessor | null;
 
   /**
    * A record of unknown devices (Devices without labels)
@@ -343,6 +349,35 @@ class AudioHelper extends EventEmitter {
   }
 
   /**
+   * Adds an {@link AudioProcessor}. If an {@link AudioProcessor} exists in the
+   * {@link AudioHelper}, the {@link AudioHelper} routes the input audio stream
+   * to the {@link AudioProcessor}.
+   *
+   * The {@link AudioHelper} guarantees that this audio stream is always
+   * the active input audio stream and will automatically
+   * update whenever the user's preference changes.
+   *
+   * Only one {@link AudioProcessor} can be added at this time.
+   * @param processor
+   */
+  addProcessor(processor: AudioProcessor): void {
+    if (this._processor) {
+      throw new NotSupportedError('Adding multiple AudioProcessors is not supported at this time.');
+    }
+
+    if (!processor) {
+      throw new InvalidArgumentError('Missing AudioProcessor argument.');
+    }
+
+    if (typeof processor.createProcessedStream !== 'function' ||
+    typeof processor.destroyProcessedStream !== 'function') {
+      throw new InvalidArgumentError('Missing createProcessedStream or destroyProcessedStream.');
+    }
+
+    this._processor = processor;
+  }
+
+  /**
    * Enable or disable the disconnect sound.
    * @param doEnable Passing `true` will enable the sound and `false` will disable the sound.
    * Not passing this parameter will not alter the enable-status of the sound.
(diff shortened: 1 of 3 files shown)


lib/twilio/audiohelper.ts
```diff
@@ -342,6 +348,35 @@ class AudioHelper extends EventEmitter {
     });
   }
 
+  /**
+   * Adds an {@link AudioProcessor}. If an {@link AudioProcessor} exists in the
```
mhuynh5757, line 352: In my opinion, the wording here seems to imply "If an AudioProcess (already) exists in the..." due to grammatical structure. My gut instinct reading this is that the functionality only works on the second or further invocation of the function, which I am assuming is not true. I'm trying to think of how I would word this... maybe something like

```
Adds an AudioProcessor.

Once the AudioProcessor contains at least one AudioProcessor, the AudioHelper routes the input audio stream to the AudioProcessor.
```

mhuynh5757: I understand that currently we do not support multiple AudioProcessors, so maybe you'd want to reword that.

lib/twilio/audiohelper.ts
```diff
@@ -342,6 +348,35 @@ class AudioHelper extends EventEmitter {
...
+   *
+   * Only one {@link AudioProcessor} can be added at this time.
+   * @param processor
+   */
+  addProcessor(processor: AudioProcessor): void {
+    if (this._processor) {
+      throw new NotSupportedError('Adding multiple AudioProcessors is not supported at this time.');
+    }
+
+    if (!processor) {
```
mhuynh5757, line 368: Do we want to do something a little stricter here like `typeof processor != 'object'`?

lib/twilio/audiohelper.ts
```diff
@@ -342,6 +348,35 @@ class AudioHelper extends EventEmitter {
...
+      throw new InvalidArgumentError('Missing AudioProcessor argument.');
+    }
+
+    if (typeof processor.createProcessedStream !== 'function' ||
+    typeof processor.destroyProcessedStream !== 'function') {
+      throw new InvalidArgumentError('Missing createProcessedStream or destroyProcessedStream.');
+    }
```
mhuynh5757, line 375:
```suggestion
    if (
      typeof processor.createProcessedStream !== 'function' ||
      typeof processor.destroyProcessedStream !== 'function'
    ) {
      throw new InvalidArgumentError('Missing createProcessedStream or destroyProcessedStream.');
    }
```
Nitpick

mhuynh5757 (commented):
Left a few comments, looks good!

lib/twilio/audiohelper.ts
manjeshbhargav, line 368: This change will make sure a value of `null` will not trigger this error:
```suggestion
    if (typeof processor === 'undefined') {
```

lib/twilio/audiohelper.ts
```diff
@@ -342,6 +348,35 @@ class AudioHelper extends EventEmitter {
...
+      throw new NotSupportedError('Adding multiple AudioProcessors is not supported at this time.');
+    }
+
+    if (!processor) {
+      throw new InvalidArgumentError('Missing AudioProcessor argument.');
+    }
+
+    if (typeof processor.createProcessedStream !== 'function' ||
+    typeof processor.destroyProcessedStream !== 'function') {
+      throw new InvalidArgumentError('Missing createProcessedStream or destroyProcessedStream.');
```
manjeshbhargav, line 374:
```suggestion
    if (typeof processor.createProcessedStream !== 'function') {
      throw new InvalidArgumentError('Missing createProcessedStream() method.');
    }

    if (typeof processor.destroyProcessedStream !== 'function') {
      throw new InvalidArgumentError('Missing destroyProcessedStream() method.');
```

manjeshbhargav (approved):

lib/twilio/audiohelper.ts
charliesantos: I'm not sure what you mean by "the functionality only works on the second or further invocation of the function". This was reviewed with the whole team. And another one with you in the IDL. I'm just wondering why the confusion now? Can you elaborate?

lib/twilio/audiohelper.ts
charliesantos: Good point. But @manjeshbhargav 's suggestion seems better. It accounts for null.

charliesantos: Testing it more, your suggestion is better but with a check on null. Will update.

charliesantos: We don't want the user to set this to null using this method. I'll check for object and null instead.

mhuynh5757 (approved):

PR #213, opened 2023-11-07
VBLOCKS-2032 | AudioProcessor insights and reusing AudioHelper

charliesantos:
**Contributing to Twilio**

## Pull Request Details

- AudioProcessor Insights. See https://issues.corp.twilio.com/browse/VBLOCKS-2032.
- Using observer pattern with separate observable to separate public events (device.audio) vs internal (audio processor).
- Reusing audio helper instance, fixing a bug where calling device.updateOptions causes audio helper memory leak.
- Tests are no included. They are part of a separate ticket.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/audiohelper.ts
+++ b/lib/twilio/audiohelper.ts
@@ -4,6 +4,7 @@
  */
 import { EventEmitter } from 'events';
 import AudioProcessor from './audioprocessor';
+import { AudioProcessorEventObserver } from './audioprocessoreventobserver';
 import Device from './device';
 import { InvalidArgumentError, NotSupportedError } from './errors';
 import Log from './log';
@@ -89,6 +90,11 @@ class AudioHelper extends EventEmitter {
    * An AudioContext to use.
    */
   private _audioContext?: AudioContext;
+
+  /**
+   * The AudioProcessorEventObserver instance to use
+   */
+  private _audioProcessorEventObserver: AudioProcessorEventObserver;
 
   /**
    * The audio stream of the default device.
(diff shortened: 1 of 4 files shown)


lib/twilio/audiohelper.ts
```diff
@@ -399,6 +410,19 @@ class AudioHelper extends EventEmitter {
     });
   }
 
+  /**
+   * Update AudioHelper options that can be changed by the user
+   * @private
+   */
+  _updateUserOptions(options: AudioHelper.Options): void {
```
mhuynh5757, line 417:
```suggestion
  private _updateUserOptions(options: AudioHelper.Options): void {
```

lib/twilio/audiohelper.ts
```diff
@@ -262,11 +268,16 @@ class AudioHelper extends EventEmitter {
   }
 
   /**
-   * Current state of the enabled sounds
+   * Destroy this AudioHelper instance
    * @private
    */
-  _getEnabledSounds(): Record<Device.ToggleableSound, boolean> {
-    return this._enabledSounds;
+  _destroy(): void {
```
mhuynh5757, line 274:
```suggestion
  private _destroy(): void {
```

lib/twilio/device.ts
```diff
@@ -887,8 +890,7 @@ class Device extends EventEmitter {
    */
   private _destroyAudioHelper() {
     if (!this._audio) { return; }
-
-    this._audio.removeAllListeners();
+    this._audio._destroy();
```
mhuynh5757, line 893: Is this why we didn't actually mark `_destroy` as a private function? Should we just remove the `_`?

lib/twilio/audiohelper.ts
charliesantos: This is public internal. Meaning, customers cannot use it (not visible in public API), but internal classes can. See device.

lib/twilio/audiohelper.ts
charliesantos: This is public internal. Meaning, customers cannot use it (not visible in public API), but internal classes can. See device.

lib/twilio/device.ts
charliesantos: Correct. We cannot remove `_` because that implies it's a customer API.

mhuynh5757 (approved):

PR #218, opened 2023-11-28
Update mediaHandler.onopen to use mediaHandler.isMuted to mute/unmute

zyzmoz:
**Contributing to Twilio**

## Pull Request Details

Updated `mediaHandler.onopen` to use its `isMuted` property when muting/unmuting the call

### Description

It is possible to mute/unmute the call prior to the media opening, causing divergence between the actions performed (mute/unmute) and the actual state (which would be always unmuted)

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/tests/unit/call.ts
+++ b/tests/unit/call.ts
@@ -1446,6 +1446,19 @@ describe('Call', function() {
             sinon.assert.calledWith(conn.mute as SinonSpy, false);
           });
 
+          context('when this call is muted', () => {
+            beforeEach(() => {
+              mediaHandler.isMuted = true;
+            });
+
+            it(`should call call.mute(true)`, () => {
+              conn.mute = sinon.spy();
+              mediaHandler.onopen();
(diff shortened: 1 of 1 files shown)


charliesantos:
@zyzmoz thank you for submitting this PR. Can you please provide more details regarding the issue this PR is trying to solve? Such as steps to reproduce, expected and actual behavior.

zyzmoz:
So, when you make a call, you can mute or unmute that before the call is answered. If that call is muted and the other end answers, then the call will be unmuted by default. The changes I made solved that issue by using the `mediaHandler` state to mute/unmute the call when `mediaHandler.onopen` is triggered. Also, as the `mediaHandler.isMuted` default is `false`, the previous behaviour should be covered.

###  Steps to reproduce

1. Make a call
2. Mute on the Caller end without answering the call
3. Callee answers the call

**Result:** Call gets unmuted

PR #221, opened 2023-11-29
VBLOCKS-1850 Add and improve debug logs

charliesantos:
**Contributing to Twilio**

## Pull Request Details

See VBLOCKS-1850

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/device.ts
+++ b/lib/twilio/device.ts
@@ -377,7 +377,7 @@ class Device extends EventEmitter {
   /**
    * An instance of Logger to use.
    */
-  private _log: Log = Log.getInstance();
+  private _log: Log = new Log('Device');
 
   /**
    * Network related information
@@ -467,6 +467,10 @@ class Device extends EventEmitter {
   constructor(token: string, options: Device.Options = { }) {
     super();
 
+    // Setup loglevel asap to avoid missed logs
+    this._setupLoglevel(options.logLevel);
+    this._logOptions('constructor', options);
+
     this.updateToken(token);
 
     if (isLegacyEdge()) {
@@ -554,6 +558,7 @@ class Device extends EventEmitter {
    * @param options
    */
   async connect(options: Device.ConnectOptions = { }): Promise<Call> {
+    this._log.debug('.connect', JSON.stringify(options && options.params || {}), options);
     this._throwIfDestroyed();
 
     if (this._activeCall) {
@@ -592,6 +597,7 @@ class Device extends EventEmitter {
    * Destroy the {@link Device}, freeing references to be garbage collected.
    */
   destroy(): void {
+    this._log.debug('.destroy');
     this.disconnectAll();
     this._stopRegistrationTimer();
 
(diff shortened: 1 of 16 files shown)


lib/twilio/device.ts
```diff
@@ -1367,6 +1423,15 @@ class Device extends EventEmitter {
     });
   }
 
+  /**
+   * Setup logger's loglevel
+   */
+  private _setupLoglevel(logLevel?: LogLevelDesc): void {
+    const level = typeof logLevel === 'number' ? logLevel : LogLevels.ERROR;
```
mhuynh5757, line 1430: This limits users to only be able to use numbers, but LogLevelDesc also has strings such as `error` and `debug`.

lib/twilio/eventpublisher.ts
```diff
@@ -58,7 +59,7 @@ class EventPublisher extends EventEmitter {
         get() { return isEnabled; },
         set(_isEnabled) { isEnabled = _isEnabled; },
       },
-      _log: { value: options.log },
+      _log: { value: new Log('EventPublisher') },
```
mhuynh5757, line 62: Should we still allow an option override here? `options.log || new Log`?

lib/twilio/eventpublisher.ts
```diff
@@ -117,6 +126,11 @@ EventPublisher.prototype._post = function _post(endpointName, level, group, name
     event.publisher_metadata = this.metadata;
   }
 
+  endpointName === 'EndpointEvents' && this._log.debug(
```
mhuynh5757, line 129: Should we just use a traditional `if` statement?

lib/twilio/preflight/preflight.ts
```diff
@@ -196,12 +202,43 @@ export class PreflightTest extends EventEmitter {
       fileInputStream: this._options.fakeMicInput ?
         this._getStreamFromFile() : undefined,
     });
+
+    // Device sets the loglevel so start logging after initializing the device.
+    // Then selectively log options that users can modify.
+    const userOptions = [
```
mhuynh5757, line 208: It seems like we do this kind of thing in several places. Should we extract options behavior this into a separate module? Maybe in a future ticket?

lib/twilio/device.ts
charliesantos: Good catch! This seems like an existing bug (I only moved this logic to its own method). I'll create another ticket to not block this PR. Let me know if you don't agree.

lib/twilio/eventpublisher.ts
charliesantos: We don't need to do that anymore. Actually, we shouldn't do that anymore because we want each class to create their own log instance.

lib/twilio/eventpublisher.ts
charliesantos: Yup, old habits. Actually I already have this fixed locally. Will push up after I fix the tests.

lib/twilio/preflight/preflight.ts
mhuynh5757: I think there's a lot of room for improving our options DX in general, I think the typing needs a bit of cleanup and perhaps the options overrides should be moved into their own module. Probably out of scope for this ticket. We can worry about it later I agree.

mhuynh5757 (approved):

PR #252, opened 2024-04-09
[VBLOCKS-1009] feat: publish insights events

kpchoy:
**Contributing to Twilio**

### JIRA link(s):

- [VBLOCKS-1009](https://issues.corp.twilio.com/browse/VBLOCKS-1009)

### Description

Add insights events to callmessage sent/received/error

#### Errors
Screenshot 2024-04-09 at 11 19 07 AM

Screenshot 2024-04-08 at 4 43 33 PM

#### Sent
Screenshot 2024-04-08 at 3 53 04 PM

#### Received 
Screenshot 2024-04-08 at 3 52 49 PM

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/call.ts
+++ b/lib/twilio/call.ts
@@ -1356,8 +1356,7 @@ class Call extends EventEmitter {
       messageType: messagetype,
       voiceEventSid: voiceeventsid,
     };
-    this._publisher.info('call-message', 'user-defined-message', {
-      call_sid: callsid,
+    this._publisher.info('call-message', messagetype, {
       content_type: contenttype,
       event_type: 'received',
       voice_event_sid: voiceeventsid,
@@ -1378,9 +1377,8 @@ class Call extends EventEmitter {
     }
     const message = this._messages.get(voiceEventSid);
     this._messages.delete(voiceEventSid);
-    this._publisher.info('call-message', 'user-defined-message', {
-      call_sid: this.parameters.CallSid,
-      content_type: this._messages.get(voiceEventSid)?.contentType,
+    this._publisher.info('call-message', message?.messageType, {
+      content_type: message?.contentType,
       event_type: 'sent',
       voice_event_sid: voiceEventSid,
     }, this);
@@ -1433,7 +1431,7 @@ class Call extends EventEmitter {
    * Called when an 'error' event is received from the signaling stream.
    */
   private _onSignalingError = (payload: Record<string, any>): void => {
-    const { callsid, voiceeventsid } = payload;
+    const { callsid, voiceeventsid, error } = payload;
     if (this.parameters.CallSid !== callsid) {
       this._log.warn(`Received an error from a different callsid: ${callsid}`);
       return;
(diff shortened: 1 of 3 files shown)


lib/twilio/call.ts
```diff
@@ -1356,6 +1356,12 @@ class Call extends EventEmitter {
       messageType: messagetype,
       voiceEventSid: voiceeventsid,
     };
+    this._publisher.info('call-message', 'user-defined-message', {
```
charliesantos, line 1359: 'user-defined-message' shouldn't be hardcoded. let's use the `messagetype` variable. update onMessageSent as well.

lib/twilio/call.ts
```diff
@@ -1356,6 +1356,12 @@ class Call extends EventEmitter {
...
+      call_sid: callsid,
```
charliesantos, line 1360: I think call_sid is auto populated from the call object. Can you try removing it and see if it still gets sent? update onMessageSent as well.

lib/twilio/device.ts
```diff
@@ -1223,6 +1223,16 @@ class Device extends EventEmitter {
...
+      } else if (code === 31209 ) {
+        this._publisher.error('call-message', 'error', {
+          code,
+          message: customMessage,
+        }, this._activeCall);
+      } else if (code === 31100 ) {
+        this._publisher.error('call-message', 'error', {
+          code,
+          message: customMessage,
+        }, this._activeCall);
```
charliesantos, line 1235: If we only check for specific error codes, we will miss other future errors unless we update the code. Let's make this dynamic. I suggest use this [if block](https://github.com/twilio/twilio-voice.js/blob/feature/eventing_1.5/lib/twilio/call.ts#L1429). We already check in there if the error is related to call message. You can use the payload object to read the information you need to send to insights. Also, we need to send `voice_event_sid` as well. It's missing in your code above. See schema here https://code.hq.twilio.com/client/rtc-insights-spec/blob/master/schema/voice/event/call-message/error/index.json

charliesantos (commented):
Left a review @kpchoy . Also, can you please add unit tests?

lib/twilio/call.ts
kpchoy: I thought we got rid of the enum for 'user-defined-message'? Should I add the enum back?

lib/twilio/device.ts
kpchoy: The schema has it as "not required". It looks like from the Kibana logs it is getting auto populated. Do you still want me to add it in?

lib/twilio/call.ts
charliesantos: No. There's a `messagetype` variable at the top. We should use that.

lib/twilio/device.ts
charliesantos: Are you talking about voice event sid or callsid?

kpchoy: `voice_event_sid` it is not in the "required" section: https://code.hq.twilio.com/client/rtc-insights-spec/blob/master/schema/voice/event/call-message/error/index.json#L35

tests/unit/call.ts
```diff
@@ -2085,6 +2092,33 @@ describe('Call', function() {
...
+      });
+
+      it('should publish a user-defined-message error event', () => {
+        const payloadPromise = new Promise((resolve, reject) => {
+          conn.on('error', reject);
+        });
+        payloadPromise.then(() => {
+          sinon.assert.calledWith(publisher.error, 'user-defined-message', 'error');
+        });
+      });
```
charliesantos, line 2121: Can you also test the correct payload is published?

charliesantos (changes requested):

charliesantos (approved):

PR #253, opened 2024-04-11
VBLOCKS-2929 VBLOCKS-2927 Manual reconnection API

charliesantos:
**Contributing to Twilio**

## Pull Request Details

See https://docs.google.com/document/d/1QMwkNXlJYSANksP0BrlJ_JAbJA86x85ZKKQ8656GnUw/edit#heading=h.woffqrxm0odm

// TODO (different tickets)
- changelog
- tests

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/device.ts
+++ b/lib/twilio/device.ts
@@ -563,16 +563,19 @@ class Device extends EventEmitter {
       throw new InvalidStateError('A Call is already active');
     }
 
-    let callSid, customParameters, parameters, signalingReconnectToken;
+    let customParameters, parameters, signalingReconnectToken;
     if (options.connectToken) {
       try {
-        const parts = JSON.parse(atob(options.connectToken));
-        callSid = parts.callSid;
-        customParameters = parts.customParameters;
-        parameters = parts.parameters;
-        signalingReconnectToken = parts.signalingReconnectToken;
+        const connectTokenParts = JSON.parse(atob(options.connectToken));
+        customParameters = connectTokenParts.customParameters;
+        parameters = connectTokenParts.parameters;
+        signalingReconnectToken = connectTokenParts.signalingReconnectToken;
       } catch {
         throw new InvalidArgumentError('Cannot parse connectToken');
+      }
+
+      if (!parameters || !parameters.CallSid || !signalingReconnectToken) {
+        throw new InvalidArgumentError('Invalid connectToken');
       }
     }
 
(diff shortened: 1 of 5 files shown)


lib/twilio/device.ts
```diff
@@ -559,22 +557,45 @@ class Device extends EventEmitter {
...
-      },
-    );
+    let callSid, customParameters, parameters, signalingReconnectToken;
+    if (options.connectToken) {
+      try {
+        const parts = JSON.parse(atob(options.connectToken));
+        callSid = parts.callSid;
+        customParameters = parts.customParameters;
+        parameters = parts.parameters;
+        signalingReconnectToken = parts.signalingReconnectToken;
```
mhuynh5757, line 573: Do you think we should add checks here for `typeof part === 'string' && part.length > 0`?

lib/twilio/device.ts
```diff
@@ -559,22 +557,45 @@ class Device extends EventEmitter {
...
-        enableImprovedSignalingErrorPrecision:
-          !!this._options.enableImprovedSignalingErrorPrecision,
-        rtcConfiguration: options.rtcConfiguration,
-        voiceEventSidGenerator: this._options.voiceEventSidGenerator,
-      },
-    );
+    let callSid, customParameters, parameters, signalingReconnectToken;
+    if (options.connectToken) {
+      try {
+        const parts = JSON.parse(atob(options.connectToken));
```
kpchoy, line 569: `parts` seems too general of a variable name, how about `connectTokenParts` or `connectTokenParams`?

lib/twilio/device.ts
charliesantos: Not 100% sure what you mean. Can you add the code snippet?

mhuynh5757:
```typescript
callSid = parts.callSid;
if (typeof callSid !== 'string' || callSid.length === 0) {
  throw...
}
```
For each of the parts.

charliesantos: Ok. I added validation only for the required values. Also, I'm only checking for truthy values instead of checking every single type and length for each parameter. Keeps the code a bit simpler. If the user re-encodes the connectToken with bad types which is highly unlikely, we fail anyway on the server.

kpchoy (approved):

mhuynh5757 (approved):

PR #262, opened 2024-04-30
Feature/chrome ext mv3

charliesantos:
This is a merge PR for the Chrome MV3 and incoming call forwarding feature.


mhuynh5757 (approved):

mhuynh5757:
I was not _very_ thorough considering that this is a merge PR. Please let me know if I should go through the PR with a closer inspection.

charliesantos:
> I was not _very_ thorough considering that this is a merge PR. Please let me know if I should go through the PR with a closer inspection.

No need for another thorough review since all the merge commits here has already been reviewed.

PR #279, opened 2024-07-11
Fix CSP in README.md

benweissmann:
Per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/Sources#sources -- the correct format for protocol source values has a trailing colon.

**Contributing to Twilio**


charliesantos:
Thanks @benweissmann .

PR #280, opened 2024-07-17
Handle when `rtcIceCandidateStatsReport` is undefined

phi-line:
**Contributing to Twilio**

### JIRA link(s):

### Description

In the case where rtcIceCandidateStatsReport is never set, we cannot read from one of its properties. This change ensures that a default value is set when rtcIceCandidateStatsReport is undefined.

### Before review
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/preflight/preflight.ts
+++ b/lib/twilio/preflight/preflight.ts
@@ -292,7 +292,7 @@ export class PreflightTest extends EventEmitter {
     const report: PreflightTest.Report = {
       callSid: this._callSid,
       edge: this._edge,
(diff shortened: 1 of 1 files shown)


charliesantos:
@phi-line thanks for submitting this PR. It's simple and I understand why this is needed. We'll take a closer look and make sure there are no side effects.

charliesantos:
Thanks @phi-line . Seems like a good fix. We're currently working on this and expect a release in the next few days.

Issue #228, opened 2023-12-12
[BUG] Error while calling setInputDevice when twilio sdk was instantiated with microphone permissions not granted


PriscilaAlves:
- [x] I have verified that the issue occurs with the latest twilio.js release and is not marked as a known issue in the [CHANGELOG.md](https://github.com/twilio/twilio-voice.js/blob/master/CHANGELOG.md).
 - [x] I reviewed the [Common Issues](https://github.com/twilio/twilio-voice.js/blob/master/COMMON_ISSUES.md) and open GitHub issues and verified that this report represents a potentially new issue.
 - [ ] I verified that the [Quickstart](https://github.com/TwilioDevEd/voice-javascript-sdk-quickstart-node) application works in my environment.
 - [x] I am not sharing any [Personally Identifiable Information (PII)](https://www.twilio.com/docs/glossary/what-is-personally-identifiable-information-pii)
       or sensitive account information (API keys, credentials, etc.) when reporting this issue.

**Code to reproduce the issue:**

When the app is opened and Twilio device is initialized, ensure you haven't granted microphone permissions yet.
Then, grant permissions to microphone and only after call `setInputDevice`
```js
this.twilioDevice = new Device(token, options)

// grant permissions to microphone in the browser
try {
 await this.twilioDevice.audio.setInputDevice('default')
} catch (e) {
 console.log(e)
}
```

**Expected behavior:**

It should succeed to set the input device since the permissions are already granted

**Actual behavior:**

It fails to set input device with the error:
`InvalidArgumentError: Device not found: default`

**Software versions:**

 - [x] Browser(s): Chrome Version 119.0.6045.199 (Official Build) (arm64)
 - [x] Operating System: Mac OS
 - [x] twilio.js: 2.8.0
 - [ ] Third-party libraries (e.g., Angular, React, etc.):

charliesantos:
@PriscilaAlves thanks for reaching out! This happens because the device ID that you're passing to setInputDevice is not detected. To properly update the internal list of device IDs, you need to grant permissions before initializing the device. For example:
```ts
// ask permission
const stream = await navigator.mediaDevices.getUserMedia({ audio: true });
stream.getTracks().forEach(track => track.stop());

this.twilioDevice = new Device(token, options);
await this.twilioDevice.audio.setInputDevice('default')
```

PriscilaAlves:
@charliesantos Thanks for the response.
[...]

PR #283, opened 2024-08-01
[VBLOCKS-3256] Feat: listen for navigator permissions changes

kpchoy:
**Contributing to Twilio**

### JIRA link(s):

- [VBLOCKS-3256](https://twilio-engineering.atlassian.net/browse/VBLOCKS-3256)

### Description

Reference issue: https://github.com/twilio/twilio-voice.js/issues/228
- allow user to instantiate the Twilio Device after the permissions are set
- listen to microphone permission changes

## Burndown

Tested:
* [x] Chrome
* [x] Safari
* [x] Edge
* [ ] Firefox - Firefox shows error, but is able to make outgoing call. Microphone permissions not supported in firefox https://github.com/mozilla/standards-positions/issues/19#issuecomment-370158947 

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/audiohelper.ts
+++ b/lib/twilio/audiohelper.ts
@@ -275,6 +275,16 @@ class AudioHelper extends EventEmitter {
     if (isEnumerationSupported) {
       this._initializeEnumeration();
     }
+
+    navigator.permissions.query({ name: 'microphone' as PermissionName }).then((microphonePermissionStatus) => {
+      if (microphonePermissionStatus.state !== 'granted') {
+        const handleStateChange = () => {
+          this._updateAvailableDevices();
+          microphonePermissionStatus.removeEventListener('change', this._updateAvailableDevices);
+        };
+        microphonePermissionStatus.addEventListener('change', handleStateChange);
+      }
+    }).catch((reason) => this._log.warn(`Warning: unable to listen for microphone permission changes. ${reason}`));
   }
 
   /**
--- a/lib/twilio/audiohelper.ts
+++ b/lib/twilio/audiohelper.ts
@@ -164,9 +164,19 @@ class AudioHelper extends EventEmitter {
   private _mediaDevices: AudioHelper.MediaDevicesLike | null;
 
   /**
+   * The microphone permission status
+   */
+  private _microphonePermissionStatus: PermissionStatus;
+
+  /**
    * Called with the new input stream when the active input is changed.
    */
   private _onActiveInputChanged: (stream: MediaStream | null) => Promise<void>;
+
+  /**
+   * Handler for microphone permission status change
+   */
+  private _onMicrophonePermissionStatusChanged: () => void;
 
   /**
    * Internal reference to the processed stream
@@ -276,13 +286,17 @@ class AudioHelper extends EventEmitter {
       this._initializeEnumeration();
     }
 
-    navigator.permissions.query({ name: 'microphone' as PermissionName }).then((microphonePermissionStatus) => {
+    // NOTE (kchoy): Currently microphone permissions are not supported in firefox.
+    // https://github.com/mozilla/standards-positions/issues/19#issuecomment-370158947
+    navigator.permissions.query({ name: 'microphone' }).then((microphonePermissionStatus) => {
       if (microphonePermissionStatus.state !== 'granted') {
         const handleStateChange = () => {
           this._updateAvailableDevices();
           microphonePermissionStatus.removeEventListener('change', this._updateAvailableDevices);
         };
         microphonePermissionStatus.addEventListener('change', handleStateChange);
+        this._microphonePermissionStatus = microphonePermissionStatus;
+        this._onMicrophonePermissionStatusChanged = handleStateChange;
       }
     }).catch((reason) => this._log.warn(`Warning: unable to listen for microphone permission changes. ${reason}`));
   }
@@ -297,6 +311,7 @@ class AudioHelper extends EventEmitter {
     this._destroyProcessedStream();
     this._maybeStopPollingVolume();
     this.removeAllListeners();
+    this._stopMicrophonePermissionListener();
     this._unbind();
   }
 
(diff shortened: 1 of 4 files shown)


lib/twilio/audiohelper.ts
```diff
@@ -275,6 +275,16 @@ class AudioHelper extends EventEmitter {
     if (isEnumerationSupported) {
       this._initializeEnumeration();
     }
+
```
charliesantos, line 278: Let's add a comment here about firefox not working and add a link to the firefox issue.

lib/twilio/audiohelper.ts
```diff
@@ -275,6 +275,16 @@ class AudioHelper extends EventEmitter {
...
+    navigator.permissions.query({ name: 'microphone' as PermissionName }).then((microphonePermissionStatus) => {
+      if (microphonePermissionStatus.state !== 'granted') {
+        const handleStateChange = () => {
+          this._updateAvailableDevices();
+          microphonePermissionStatus.removeEventListener('change', this._updateAvailableDevices);
```
charliesantos, line 283:
```suggestion
          microphonePermissionStatus.removeEventListener('change', handleStateChange);
```

charliesantos: We also need to call removeEventListener in the `_destroy()` function. Or else, this will cause a memory leak if the permission never changes after destroying the audio helper. To do this, you might want to save a reference to the microphonePermissionStatus and the change handler: `this._microphonePermissionStatus` and `this._onMicrophonePermissionStatusChanged`.

charliesantos (changes requested):
Left some comments. Also
- add changelog
- update from master

lib/twilio/audiohelper.ts
```diff
@@ -275,6 +275,16 @@ class AudioHelper extends EventEmitter {
     if (isEnumerationSupported) {
       this._initializeEnumeration();
     }
+
+    navigator.permissions.query({ name: 'microphone' as PermissionName }).then((microphonePermissionStatus) => {
```
mhuynh5757, line 279: Do we need to cast `'microphone' as PermissionName` here? I don't believe so from my testing unless I'm missing something.

tests/audiohelper.js
```diff
@@ -124,6 +156,24 @@ describe('AudioHelper', () => {
...
+        let microphonePermissionStatus = {
+          state: 'prompt',
+          addEventListener: (eventName, listener) => {
+            if (eventName === 'change') {
+              listener()
+            }
+          }
+        };
+        navigator.permissions.query = () => Promise.resolve(microphonePermissionStatus);
+        audio = new AudioHelper(onActiveOutputsChanged, onActiveInputChanged, {});
```
mhuynh5757, line 171: It seems like this test might not going through SDK codepaths, unless my understanding is incorrect. If you remove these two lines, I don't think much changes? It seems like you're creating the `microphonePermissionStatus` object above and then lines `172~173` test that code path?

mhuynh5757 (changes requested):

lib/twilio/audiohelper.ts
kpchoy: My VScode is giving my a type error, but it looks like it compiles fine without the typecast 👍

tests/audiohelper.js
kpchoy: I see your point :( . I am having a hard time testing this feature correctly. Any suggestions?
cc: @charliesantos

lib/twilio/audiohelper.ts
```diff
@@ -382,6 +407,16 @@ class AudioHelper extends EventEmitter {
     }
   }
 
+  /**
+   * Remove event listener for microphone permissions
+   * @private
+   */
+  _stopMicrophonePermissionListener(): void {
```
charliesantos, line 414:
```suggestion
  /**
   * Remove event listener for microphone permissions
   */
  private _stopMicrophonePermissionListener(): void {
```
You can set this to private (instead of internal public) since it's not used outside of this file.

tests/audiohelper.js
charliesantos: Let's get the implementation approved then I'll review the tests 😄 
That way, we don't have to keep updating the tests while we review the PR.

charliesantos (changes requested):

lib/twilio/audiohelper.ts
```diff
@@ -275,6 +285,19 @@ class AudioHelper extends EventEmitter {
     if (isEnumerationSupported) {
       this._initializeEnumeration();
     }
+
+    // NOTE (kchoy): Currently microphone permissions are not supported in firefox.
+    // https://github.com/mozilla/standards-positions/issues/19#issuecomment-370158947
+    navigator.permissions.query({ name: 'microphone' }).then((microphonePermissionStatus) => {
+      if (microphonePermissionStatus.state !== 'granted') {
+        const handleStateChange = () => {
+          this._updateAvailableDevices();
```
charliesantos, line 294: I see that you removed the line where we unsubscribe to the change event. Should we call `this._stopMicrophonePermissionListener();` here instead?

charliesantos (changes requested):

charliesantos (changes requested):

lib/twilio/audiohelper.ts
kpchoy: oops yes will put it here as well

tests/index.ts
```diff
@@ -73,6 +73,11 @@ root.navigator = {
   },
   platform: 'platform',
   userAgent: 'userAgent',
+  permissions: {
+    query: function() {
+      return Promise.resolve('prompt');
+    }
+  }
```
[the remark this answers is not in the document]
kpchoy: If I omit this, [this device test](https://github.com/twilio/twilio-voice.js/blob/master/tests/unit/device.ts#L101) fails when `Device._setupAudioHelper` is called. I thought there were more cases where `audiohelper` is called in other files, but this seems to be the only test case that fails without the `permissions`

charliesantos (approved):
@kpchoy +1 on the implementation. Please check the tests and +1 if you approve.

kpchoy (commented):
Manual test lgtm +1

mhuynh5757 (approved):

PR #287, opened 2024-08-21
[VBLOCKS-3166] fix: setinputdevice await connect

kpchoy:
**Contributing to Twilio**

### JIRA link(s):

[VBLOCKS-3166](https://twilio-engineering.atlassian.net/browse/VBLOCKS-3166)

### Description

Calling connect()  without waiting for the promise to get resolved, then call setInputDevice() right away results into an AcquisitionFailedError.

### Before review
    * [x] Chrome
    * [x] Edge
    * [x] Safari
    * [x] Firefox
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


lib/twilio/audiohelper.ts
```diff
@@ -780,7 +788,14 @@ class AudioHelper extends EventEmitter {
...
    */
-  private _setInputDevice(deviceId: string, forceGetUserMedia: boolean): Promise<void> {
+  private async _setInputDevice(deviceId: string, forceGetUserMedia: boolean): Promise<void> {
+    const connectPromise = this._getConnectPromise();
+    if (connectPromise) {
+      this._log.debug('connectPromise detected, waiting...');
+      await connectPromise;
+      this._log.debug('connectPromise resolved');
+    }
+
```
charliesantos, line 798: If you add a default, you can change this to one line
```
await this._getConnectPromise();
```

lib/twilio/device.ts
```diff
@@ -556,62 +561,68 @@ class Device extends EventEmitter {
...
+      this._publishNetworkChange();
+      return activeCall;
+    };
 
-    activeCall.accept({ rtcConstraints: options.rtcConstraints });
-    this._publishNetworkChange();
-    return activeCall;
+    return this._connectPromise = _connect().finally(() => {
+      this._connectPromise = null;
+    });
```
charliesantos, line 625: Instead of wrapping the whole connect method, we should only wrap the async calls that are under this method. Good thing is, there's only one - makeCall. See the `await _makeCall` here. Maybe save the promise that it returns and use that to 'pause` the setInputDevice call.

charliesantos (changes requested):

lib/twilio/audiohelper.ts
```diff
@@ -123,6 +124,11 @@ class AudioHelper extends EventEmitter {
    */
   private _enumerateDevices: any;
 
+  /**
+   * Promise to wait for {@link Device.connect}.
+   */
+  private _getConnectPromise: () => Promise<Call> | null;
```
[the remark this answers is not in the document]
kpchoy: @charliesantos  when I do `() => Promise<any> = () => Promise.resolve();` , the `setInputDevice` is resolved and breaks the implementation for `inputDevicePromise`. I think this is why we need the null?

lib/twilio/audiohelper.ts
kpchoy: Do we no longer want the debug lines? I see we have debug lines https://github.com/twilio/twilio-voice.js/blob/master/lib/twilio/device.ts#L1025

charliesantos: We can add the log in the constructor where we assign the default.

lib/twilio/audiohelper.ts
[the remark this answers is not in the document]
charliesantos: Can we try adding `await _beforeSetInputDevice()` call inside the setInputDevice method? https://github.com/twilio/twilio-voice.js/blob/master/lib/twilio/audiohelper.ts#L784

lib/twilio/audiohelper.ts
kpchoy: Without the null, not sure where would be best to place this since a conditional for `options.getMakeCallPromise` will always return true. Maybe better to put the debug lines in `_makeCallPromise`?

lib/twilio/audiohelper.ts
```diff
@@ -1024,6 +1033,11 @@ namespace AudioHelper {
      */
     enumerateDevices?: any;
 
+    /**
+     * Promise to wait for {@link Device.connect}.
+     */
+    getMakeCallPromise: () => Promise<any>;
+
```
charliesantos, line 1040: Rename and make optional.
```suggestion
    /**
     * Promise to wait for before setting the input device.
     */
    beforeSetInputDevice?: () => Promise<any>;

```

lib/twilio/device.ts
```diff
@@ -697,6 +702,14 @@ class Device extends EventEmitter {
...
   }
 
+  /**
+   * Promise to wait for {@link Device.connect}.
+   * @private
+   */
+  getMakeCallPromise(): Promise<any> {
+    return this._makeCallPromise;
+  }
+
```
charliesantos, line 712: We don't need to expose this. Please remove.

lib/twilio/device.ts
```diff
@@ -380,6 +380,11 @@ class Device extends EventEmitter {
    */
   private _log: Log = new Log('Device');
 
+  /**
+   * The internal promise created when calling {@link Device.makeCall}.
+   */
+  private _makeCallPromise: Promise<any> = Promise.resolve();
```
charliesantos, line 386: This needs to be optional and should not have a default value. The value should only be present when there's a call, and shouldn't be present when there's no call.
```ts
private _makeCallPromise: Promise<any> | null = null;
```

lib/twilio/device.ts
```diff
@@ -1014,162 +1027,168 @@ class Device extends EventEmitter {
    * @param twimlParams - A flat object containing key:value pairs to be sent to the TwiML app.
    * @param options - Options to be used to instantiate the {@link Call}.
    */
-  private async _makeCall(twimlParams: Record<string, string>, options?: Call.Options, isReconnect: boolean = false): Promise<Call> {
-    if (typeof Device._isUnifiedPlanDefault === 'undefined') {
-      throw new InvalidStateError('Device has not been initialized.');
-    }
+  private _makeCall(twimlParams: Record<string, string>, options?: Call.Options, isReconnect: boolean = false): Promise<Call> {
```
charliesantos, line 1030: Instead of wrapping the whole _makeCall method, I was thinking on line 609, you can do this
```ts
this._makeCallPromise = this._makeCall(twimlParams, callOptions, isReconnect);
const activeCall = this._activeCall = await this._makeCallPromise;
```
Also, can you check if we also need this inside onSignalingInvite?

charliesantos: Then wherever we're setting `this._activeCall` to null, we should also set `this._makeCallPromise` to null.

lib/twilio/device.ts
```diff
@@ -1430,6 +1449,7 @@ class Device extends EventEmitter {
       audioContext: Device.audioContext,
       audioProcessorEventObserver: this._audioProcessorEventObserver,
       enumerateDevices: this._options.enumerateDevices,
+      getMakeCallPromise: this.getMakeCallPromise.bind(this),
```
charliesantos, line 1452: If you consider my other comments, this would become
```ts
beforeSetInputDevice: () => {
  if (this._makeCallPromise) {
    this._log.debug('beforeSetInputDevice pause detected');
    return this._makeCallPromise;
  } else {
    this._log.debug('beforeSetInputDevice pause not detected, setting default');
    return Promise.resolve();
  }
}
```

lib/twilio/audiohelper.ts
```diff
@@ -619,6 +621,11 @@ class AudioHelper extends EventEmitter {
     });
   }
 
+  /**
+   * Promise to wait for {@link Device.connect}.
```
charliesantos, line 625:
```suggestion
   * Promise to wait for before setting the input device.
```

charliesantos (changes requested):

lib/twilio/audiohelper.ts
```diff
@@ -619,6 +621,11 @@ class AudioHelper extends EventEmitter {
...
+   */
+  private _beforeSetInputDevice: () => Promise<any> = () => Promise.resolve();
```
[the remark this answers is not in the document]
kpchoy: @charliesantos I did this, but I got a linting error locally and in CI: https://app.circleci.com/pipelines/github/twilio/twilio-voice.js/11925/workflows/1c1da734-4e8a-4bfe-9a8a-570392795087/jobs/22816/parallel-runs/0/steps/0-107 . wdyt?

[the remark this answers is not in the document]
kpchoy: Got it, defaulted the function in the constructor instead of the function declaration

lib/twilio/device.ts
charliesantos: @kpchoy I think you missed the last comment above ^

tests/audiohelper.js
```diff
@@ -512,6 +520,31 @@ describe('AudioHelper', () => {
...
+        it('should wait for the beforeSetInputDevice to resolve', async () => {
+          const stub = sinon.stub();
+          audio = new AudioHelper(onActiveOutputsChanged, onActiveInputChanged, {
+            audioProcessorEventObserver: eventObserver,
+            beforeSetInputDevice: () => new Promise(res => res(stub())),
+            getUserMedia,
+            mediaDevices,
+            setSinkId: () => {}
+          });
+          await audio._setInputDevice('input');
```
charliesantos, line 532: Let's use the public API
```suggestion
          await audio.setInputDevice('input');
```

tests/audiohelper.js
```diff
@@ -512,6 +520,31 @@ describe('AudioHelper', () => {
...
+        it('should reject if beforeSetInputDevice rejects', async () => {
+          const stub = sinon.stub();
+          audio = new AudioHelper(onActiveOutputsChanged, onActiveInputChanged, {
+            audioProcessorEventObserver: eventObserver,
+            beforeSetInputDevice: () => new Promise((resolve, reject) => reject()),
+            getUserMedia,
+            mediaDevices,
+            setSinkId: () => {}
+          });
+          await assert.rejects(() => audio._setInputDevice('input'));
```
charliesantos, line 545: Let's use the public API
```suggestion
          await assert.rejects(() => audio.setInputDevice('input'));
```

tests/audiohelper.js
```diff
@@ -512,6 +520,31 @@ describe('AudioHelper', () => {
...
+            getUserMedia,
+            mediaDevices,
+            setSinkId: () => {}
+          });
+          await audio._setInputDevice('input');
+          sinon.assert.calledOnce(stub);
+        });
+        
+        it('should reject if beforeSetInputDevice rejects', async () => {
+          const stub = sinon.stub();
```
charliesantos, line 537: This is not used in this test, can we remove?

tests/audiohelper.js
```diff
@@ -8,6 +8,10 @@ function getUserMedia() {
   return Promise.resolve({ id: 'default', getTracks: () => [] });
 }
 
+function beforeSetInputDevice() {
+  return Promise.resolve();
+}
```
charliesantos, line 13: Can we remove all the default beforeSetInputDevice throughout this test file? We're already setting a default in the constructor so I think we don't have to pass this default.

lib/twilio/device.ts
kpchoy: Oops sorry @charliesantos , added that in. Please take another look.

charliesantos (changes requested):
Super close! Also please add the following device.ts tests.

Under '._setupAudioHelper()':
- should wait for makeCallPromise if there's an active call
- should use a default Promise if there's no active call

Wherever we set and unset _activeCall, we should also assert the makeCallPromise is set or unset.

charliesantos (changes requested):

tests/unit/device.ts
```diff
@@ -277,12 +277,15 @@ describe('Device', function() {
...
           sinon.assert.calledOnce(stub);
           assert(!!device['_activeCall']);
+          assert(!!device['_makeCallPromise']);
         });
 
         it('should reject if inputDevicePromise rejects', async () => {
           device.audio!._getInputDevicePromise = () => new Promise((resolve, reject) => reject());
           await assert.rejects(() => device.connect());
           assert(!device['_activeCall']);
+          assert.notDeepEqual(device['_makeCallPromise'], null)
```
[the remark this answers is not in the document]
charliesantos: This brings up a bug. I'm glad we discussed and caught it. If the makeCallPromise rejects, any future setIntputDevice will fail. For example, if you do the following

```ts
this.device.audio.setInputDevice('wrong-id');
this.device.connect();
this.device.audio.setInputDevice('default');
```

The 3rd line fails although the ID is correct. This happens because we're not clearing the makeCallPromise. I think to resolve this, we need to make sure the makeCallPromise internal reference is cleared after it is resolved or rejected. I'm not sure off top of my head where to do that, can you please see what we should change in this implementation to make sure this bug doesn't happen?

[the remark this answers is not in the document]
kpchoy: Revised, let me know what you think

charliesantos (approved):
+1! Please cut a RC when you get a chance :)

PR #300, opened 2024-11-19
[VBLOCKS-3750] Fix: tests navigator

kpchoy:
**Contributing to Twilio**

### JIRA link(s):

- [VBLOCKS-3750](https://twilio-engineering.atlassian.net/browse/VBLOCKS-3750)

### Description

- fix navigator tests node22
- [globalThis.navigator](https://github.com/nodejs/node/issues/39540#issuecomment-1770612761) added in Node v22
- Tested locally node v22.8.0

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/tests/audiohelper.js
+++ b/tests/audiohelper.js
@@ -70,6 +70,7 @@ describe('AudioHelper', () => {
     let availableDevices;
     let handlers;
     let mediaDevices;
+    let oldMediaDevices;
 
     beforeEach(() => {
       eventObserver = new AudioProcessorEventObserver();
@@ -101,10 +102,15 @@ describe('AudioHelper', () => {
         setSinkId: () => {}
       });
       
+      oldMediaDevices = navigator.mediaDevices;
       navigator.mediaDevices = {
         enumerateDevices() { return Promise.resolve([]) },
       };
     });
+
+    afterEach(() => {
+      navigator.mediaDevices = oldMediaDevices;
+    })
 
     describe('constructor', () => {
       it('should set .isOutputSelectionSupported to true', () => {
--- a/tests/audiohelper.js
+++ b/tests/audiohelper.js
@@ -100,6 +100,10 @@ describe('AudioHelper', () => {
         mediaDevices,
         setSinkId: () => {}
       });
+      
+      navigator.mediaDevices = {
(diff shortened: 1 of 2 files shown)


tests/audiohelper.js
```diff
@@ -100,6 +100,10 @@ describe('AudioHelper', () => {
         mediaDevices,
         setSinkId: () => {}
       });
+      
+      navigator.mediaDevices = {
+        enumerateDevices() { return Promise.resolve([]) },
+      };
```
charliesantos, line 106: Since we're setting this in beforeEach, should we save the original, then restore it in afterEach?

tests/index.ts
```diff
@@ -68,7 +68,7 @@ root.requestAnimationFrame = () => { };
 
 root.MediaStream = () => { };
 
-root.navigator = {
+root.navigator ??= {
```
charliesantos, line 71: Can you provide an explanation why this is needed? It's great if you have a link/source.

charliesantos (changes requested):

tests/index.ts
kpchoy: Yes, looks like [globalThis.navigator](https://github.com/nodejs/node/issues/39540#issuecomment-1770612761) added in Node v22. Hence, the error `Cannot set property navigator of #<object> which has only a getter`.

charliesantos: If I remove lines 71 - 83, the tests doesn't even fail. Does this mean we don't need this anymore?

kpchoy: @charliesantos yes, but it will fail on Node v20, and I though we still want tests to pass on v20 since we use it for "stable" docker containers.

charliesantos: Good call. In this case, let's add a comment here to prevent confusion
>// This only applies to Node v21 and earlier, as v22 already includes globalThis.navigator.

charliesantos (approved):
+1 for me. Let's wait for a bit in case @mhuynh5757 and @luisrivas wants to quickly check.

mhuynh5757 (approved):

PR #321, opened 2025-03-20
Vblocks 4237 eaccess permission issues

kpchoy:
**Contributing to Twilio**

### Description

- Eaccess permission issues
- merges into working branch
- https://docs.npmjs.com/resolving-eacces-permissions-errors-when-installing-packages-globally 

### Before review
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,10 +108,13 @@ commands:
         type: boolean
         default: false
     steps:
+      - run:
+          name: Set npm default directory
+          command: bash scripts/set-npm-default-directory.sh
       - build
       - run:
           name: Install Datadog CLI
-          command: sudo npm install -g @datadog/datadog-ci
+          command: npm install -g @datadog/datadog-ci
       - when:
           condition: <<parameters.start_relay_server>>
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -109,9 +109,12 @@ commands:
         default: false
     steps:
       - build
+      - run:
+          name: Set npm default directory
+          command: bash scripts/set-npm-default-directory.sh
```
mhuynh5757, line 114: I personally don't think we need to put this in a separate bash script. Also, we can use the whole CircleCI `BASH_ENV` thing instead of a `~/.profile` file.

mhuynh5757 (changes requested):

.circleci/config.yml
charliesantos: The build process also installs npm dependencies, then we change the default directory? Should we change the default directory before everything?

kpchoy:
closing, will add to `dev-dependencies`

PR #334, opened 2025-05-19
[VBLOCKS-4563] remove 3rd party md5 module

mhuynh5757:
**Contributing to Twilio**

### JIRA link(s):

- [VBLOCKS-4563](https://twilio-engineering.atlassian.net/browse/VBLOCKS-4563)

### Description

This PR adds a vendorized version of the md5 library for internal use.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


lib/twilio/uuid.ts
```diff
@@ -4,47 +4,42 @@
...
 import { NotSupportedError } from '../twilio/errors';
 
-// If imported as an ESM module, sometimes md5 is available by accessing
-// the "default" property of the imported library.
-// @ts-ignore
-const md5 = typeof md5Lib === 'function' ? md5Lib : md5Lib.default;
-
-function generateUuid(): string {
+/**
+ * Generates a 16 character long random string where each character is from the
```
charliesantos, line 10: Why 16? The string should be 32 characters excluding the prefix.

tests/unit/uuid.ts
```diff
@@ -7,9 +7,8 @@ const root = global as any;
 
 describe('uuid util', () => {
```
charliesantos, line 8: Since it's not a uuid util anymore, can we rename all uuid references? Also, it seems character length is important for consistency. Can we add a test that checks for the length?

charliesantos (changes requested):

lib/twilio/uuid.ts
mhuynh5757: MD5 hashes are 16 characters long. All our SIDs (including CallSids, CallMessageSids, etc.) are also 16 characters long no? https://app.circleci.com/pipelines/github/twilio/twilio-voice.js/18866/workflows/ef29c2d9-01f7-425d-a615-758767923bed/jobs/33286?invite=true#step-107-1184_43

mhuynh5757: Doh, I had bytes and characters mixed up. Too used to interchanging the two from working in C-like languages. I will fix this, oopsies.

charliesantos (approved):

PR #336, opened 2025-05-27
[VBLOCKS-4686] Remove rtcpeerconnectionshim

mhuynh5757:
**Contributing to Twilio**

### JIRA link(s):

- [VBLOCKS-4686](https://twilio-engineering.atlassian.net/browse/VBLOCKS-4686)

### Description

This PR removes the `rtcpeerconnection-shim` library as it was only used for legacy edge, which our SDK has officially dropped support for.

### Before review
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/rtc/rtcpc.ts
+++ b/lib/twilio/rtc/rtcpc.ts
@@ -11,8 +11,6 @@ import Log from '../log';
 import * as util from '../util';
 import { setCodecPreferences, setMaxAverageBitrate } from './sdp';
 
-const RTCPeerConnectionShim = require('rtcpeerconnection-shim');
-
 function RTCPC(options) {
   if (typeof window === 'undefined') {
     this.log.info('No RTCPeerConnection implementation available. The window object was not found.');
@@ -22,7 +20,7 @@ function RTCPC(options) {
   if (options && options.RTCPeerConnection) {
     this.RTCPeerConnection = options.RTCPeerConnection;
   } else if (util.isLegacyEdge()) {
-    this.RTCPeerConnection = new RTCPeerConnectionShim(typeof window !== 'undefined' ? window : global);
+    this.log.error('This browser is not supported.');
   } else if (typeof window.RTCPeerConnection === 'function') {
     this.RTCPeerConnection = window.RTCPeerConnection;
   } else if (typeof window.webkitRTCPeerConnection === 'function') {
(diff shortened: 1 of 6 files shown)


lib/twilio/util.ts
```diff
@@ -62,6 +62,12 @@ function isFirefox(navigator?) {
     && /firefox|fxios/i.test(navigator.userAgent);
 }
 
+/**
+ * @privateRemarks
+ *
+ * Chromium-based Edge has a user-agent of "Edg/" where legacy Edge has a
```
kpchoy, line 68:
```suggestion
 * Chromium-based Edge has a user-agent of "Edge/" where legacy Edge has a
```

lib/twilio/util.ts
```diff
@@ -62,6 +62,12 @@ function isFirefox(navigator?) {
     && /firefox|fxios/i.test(navigator.userAgent);
 }
 
+/**
+ * @privateRemarks
```
charliesantos, line 66: Do we currently use this in our codebase? If not, can we remove it to be consistent? I understand it's probably better but it's worse if we're not being consistent.

package.json
```diff
@@ -128,10 +128,7 @@
   },
   "dependencies": {
     "@twilio/voice-errors": "1.7.0",
-    "@types/md5": "2.3.2",
     "events": "3.3.0",
-    "loglevel": "1.6.7",
-    "md5": "2.3.0",
```
charliesantos, line 134: We're removing md5 here, is it because we both missed removing it in the other PR?

lib/twilio/rtc/rtcpc.ts
```diff
@@ -22,7 +22,7 @@ function RTCPC(options) {
   if (options && options.RTCPeerConnection) {
     this.RTCPeerConnection = options.RTCPeerConnection;
   } else if (util.isLegacyEdge()) {
-    this.RTCPeerConnection = new RTCPeerConnectionShim(typeof window !== 'undefined' ? window : global);
+    this.log.info('This browser is not supported.');
```
charliesantos, line 25: We should throw an error

charliesantos (changes requested):

lib/twilio/rtc/rtcpc.ts
charliesantos: Actually let's just remove this else if block since it's EOS. Let's not make this complicated.

PR #414, opened 2026-04-06
Decouple signaling from media layer via SignalingAdapter

luisrivas:
## Pull Request Details

Introduces a `SignalingAdapter` interface and `PStreamSignalingAdapter` wrapper so alternative signaling transports (e.g. SIP) can be plugged in per-Device. `PeerConnection` is now pstream-free.

### Before review
* [x] Ready for review


lib/twilio/rtc/peerconnection.ts
```diff
@@ -827,49 +785,18 @@ PeerConnection.prototype.iceRestart = function() {
...
-  this.pstream.on('ringing', this._onAnswerOrRinging);
 
   function onOfferSuccess() {
     if (self.status !== 'closed') {
-      if (signalingReconnectToken) {
-        self.pstream.reconnect(self.version.getSDP(), self.callSid, signalingReconnectToken);
-      } else {
-        self.pstream.invite(self.version.getSDP(), self.callSid, params);
-      }
       self._setupRTCDtlsTransportListener();
```
mhuynh5757: This is something I noticed when I was doing my refactor. Putting this one before the `onOfferReady` reduces the risk of race conditions so this is a good improvement.

lib/twilio/device.ts
mhuynh5757, line 47: Since the signaling adapter is properly typed, we could change the types in this file and make things a little better.

lib/twilio/signaling/pstreamsignalingadapter.ts
```diff
@@ -0,0 +1,42 @@
...
+  'error', 'hangup', 'invite', 'message', 'offline', 'presence',
+  'ready', 'ringing', 'roster', 'transportClose', 'transportOpen',
+];
+
+export class PStreamSignalingAdapter extends EventEmitter implements SignalingAdapter {
+  private _pstream: any;
+
+  constructor(pstream: any) {
+    super();
+    this._pstream = pstream;
```
mhuynh5757, line 15: Do you think we should construct the PStream inside of this class constructor? We would have to pass options from the device down to the constructor here, which might not be relevant for other adapted signaling protocols, but it makes sense to have this adapter construct the underlying signaling transport yeah?

lib/twilio/signaling/pstreamsignalingadapter.ts
```diff
@@ -0,0 +1,42 @@
...
+  destroy(): void { this._pstream.destroy(); }
+  updatePreferredURI(uri: string | null): void { this._pstream.updatePreferredURI(uri); }
+  updateURIs(uris: string[]): void { this._pstream.updateURIs(uris); }
+
+  invite(sdp: string, callSid: string, params: string): void { this._pstream.invite(sdp, callSid, params); }
+  answer(sdp: string, callSid: string): void { this._pstream.answer(sdp, callSid); }
+  hangup(callSid: string, message?: string | null): void { this._pstream.hangup(callSid, message); }
+  reject(callSid: string): void { this._pstream.reject(callSid); }
+  reinvite(sdp: string, callSid: string): void { this._pstream.reinvite(sdp, callSid); }
+  reconnect(sdp: string, callSid: string, reconnectToken: string): void { this._pstream.reconnect(sdp, callSid, reconnectToken); }
```
mhuynh5757: Can we format these so that they look like this
```
methodName(params: ...) {
  method body
}
```
Basically, just inserting newlines.

tests/peerconnection.js
```diff
@@ -786,143 +666,125 @@ describe('PeerConnection', () => {
...
-      assert(context.pstream.on.calledWithExactly('answer', sinon.match.func));
+    it('Should set callSid on the context', () => {
+      toTest();
+      assert.equal(context.callSid, eCallSid);
     });
+  });
 
-    it('Should call onAnswer and set _answerSdp listener when pstream answer event is triggered', () => {
-      context.pstream.on.callsArgWith(1, PAYLOAD);
+  context('PeerConnection.prototype.processAnswer', () => {
```
mhuynh5757, line 706: We should consider using describe instead of context since there's a nested variable called context inside these method bodies.

lib/twilio/signaling/signalingadapter.ts
```diff
@@ -0,0 +1,30 @@
+import { EventEmitter } from 'events';
```
kpchoy, line 1: can we move this file under a signaling folder such as `lib/twilio/signaling/signalingadapter.ts`

lib/twilio/signaling/pstreamsignalingadapter.ts
```diff
@@ -0,0 +1,42 @@
+import { EventEmitter } from 'events';
```
kpchoy, line 1: can we move this file under a signaling folder such as lib/twilio/signaling/pstreamsignalingadapter.ts

lib/twilio/signaling/signalingadapter.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export type SignalingAdapterStatus = 'disconnected' | 'connected' | 'ready' | 'offline';
+
+export interface SignalingAdapter extends EventEmitter {
+  status: SignalingAdapterStatus;
+  uri: string;
+
```
kpchoy, line 10: I think we also need `gateway` and `region`

lib/twilio/rtc/peerconnection.ts
```diff
@@ -884,7 +811,29 @@ PeerConnection.prototype.makeOutgoingCall = function(params, signalingReconnectT
 
   this.version.createOffer(this.options.maxAverageBitrate, { audio: true }, onOfferSuccess, onOfferError);
 };
-PeerConnection.prototype.answerIncomingCall = function(callSid, sdp, rtcConfiguration, onMediaStarted) {
+
+PeerConnection.prototype.processAnswer = function(sdp, onMediaStarted) {
```
kpchoy, line 821: I like the decoupling, but I realized the SIP.js SDH (SipSessionDescriptionHandler) will need finer-grained access than this method provides. SIP.js drives offer/answer via SDH.getDescription()/setDescription(), which doesn't map 1:1 to these bundled methods. The SIP path may need to call version.createOffer()/processAnswer() directly.

lib/twilio/rtc/peerconnection.ts
```diff
@@ -827,49 +785,18 @@ PeerConnection.prototype.iceRestart = function() {
   });
 };
 
-PeerConnection.prototype.makeOutgoingCall = function(params, signalingReconnectToken, callsid, rtcConfiguration, onMediaStarted) {
+PeerConnection.prototype.makeOutgoingCall = function(callsid, rtcConfiguration, onOfferReady) {
```
kpchoy, line 788: Same concern as processAnswer. This bundles media stream initialization, offer creation, and DTLS setup into one call. The SIP.js SDH needs these as separate steps: _initializeMediaStream() first, then SDH.getDescription() to create the offer. The SIP path may need to bypass this method entirely.

kpchoy (changes requested):

lib/twilio/signaling/signalingadapter.ts
luisrivas: Done!

lib/twilio/signaling/pstreamsignalingadapter.ts
luisrivas: Done!

lib/twilio/device.ts
luisrivas: Good catch! I'm improved the type inside the Device implementation to encapsulate the scope of the change

lib/twilio/signaling/signalingadapter.ts
luisrivas: Done!

lib/twilio/signaling/pstreamsignalingadapter.ts
luisrivas: Done!

tests/peerconnection.js
luisrivas: I agree. Let's address that in a separate PR to keep scope focused.

lib/twilio/signaling/signalingadapter.ts
mhuynh5757: Super surprised that integration tests were passing before we exposed these values. 😵

lib/twilio/device.ts
```diff
@@ -738,8 +739,14 @@ class Device extends EventEmitter {
...
 
     const stream = await this._streamConnectedPromise;
+    if (!stream) {
+      this._log.warn('unregister called without an active stream');
+      this._setState(Device.State.Unregistered);
+      return;
+    }
     const streamOfflinePromise = new Promise(resolve => {
       stream.on('offline', resolve);
+      stream.on('close', resolve);
```
mhuynh5757: This looks like a genuine bugfix but I think maybe we leave it out of this PR for now just so that we have _identical_ behavior to the original.

mhuynh5757 (changes requested):

lib/twilio/signaling/pstreamsignalingadapter.ts
mhuynh5757: On second thought, let's just leave it the way it is for now. We can adjust in a future PR if need be.

lib/twilio/rtc/peerconnection.ts
mhuynh5757: Thinking about this again, just so we have _identical_ behavior to the original implementation, I'd like to move the method flow back to the original way.

luisrivas:
Thanks for the follow-up commits, @mhuynh5757. I think this is in good shape to merge now.

mhuynh5757 (approved):

lib/twilio/rtc/peerconnection.ts
luisrivas: That's a valid concern. I'd recommend addressing it in a separate PR to keep this one focused.

lib/twilio/rtc/peerconnection.ts
luisrivas: That's a valid concern. I'd recommend addressing it in a separate PR to keep this one focused.

kpchoy (approved):

PR #415, opened 2026-04-07
[VBLOCKS-6367] feat: useSip feature flag

kpchoy:
**Contributing to Twilio**

## Pull Request Details

https://twilio-engineering.atlassian.net/browse/VBLOCKS-6367

### Description

useSip feature flag

### Before review
* [ ] Ready for review


luisrivas (approved):
Thanks, Kevin!

lib/twilio/device.ts
```diff
@@ -2099,6 +2118,21 @@ namespace Device {
...
+      username: string;
+      password: string;
+    };
+
+    /**
+     * The SIP WebSocket server URL to connect to when {@link Device.Options.useSip}
+     * is `true` (e.g. `'wss://sip.example.com'`).
+     */
+    sipServer?: string;
+
```
mhuynh5757, line 2186: I think we can create some breathing room for ourselves here by making these options more extensible. What do you think about typing the options like this?

```
type SIPSignalingOptions = {
  useSignalingMethod: 'sip';
  sipCredentials: { username: string; password: string };
  sipServer: string;
};

type VSPSignalingOptions = {
  useSignalingMethod: 'vsp';
  chunderw: string;
};

type Device.Options = {
  ...,
  signalingOptions: SIPSignalingOptions | VSPSignalingOptions;
};
```

The only issue I see with this is that VSP signaling options are already pre-baked into the top-level Device.Options object, albeit hidden options. So either we would include it in the signalingOptions object as an override whilst still supporting the original options members, or we don't include specifically VSPSignalingOptions for now.

mhuynh5757 (changes requested):

lib/twilio/device.ts
kpchoy: I like the idea, I also moved eventgw under VSPSignalingOptions

mhuynh5757: Insights (i.e. `eventgw` option) is mutually exclusive from the signaling methodology. We will continue to use `eventgw` regardless of the signaling option we use. Might change in the future, but for the time being, we should not put `eventgw` in `signalingOptions`.

Additionally, we still need to support `chunderw` and `eventgw` in the main device options object, as removing it would be a breaking change. It's not a public visible option we let customers know about, but other teams at Twilio use these options for internal testing tools. We'll just have to support both, or if we don't want to support both, we should remove the new `VSPSignalingOptions` type and not allow it yet.

lib/twilio/device.ts
```diff
@@ -955,8 +947,32 @@ class Device extends EventEmitter {
...
+    const chunderw = vspOpts?.chunderw;
+    return typeof chunderw === 'string' ? [chunderw]
+      : Array.isArray(chunderw) ? chunderw : null;
+  }
+
+  /**
+   * Resolves the active signaling options, defaulting to VSP if not specified.
+   * Throws {@link InvalidArgumentError} if SIP is selected but required fields are missing.
+   */
+  private _resolveSignalingOptions(): Device.SIPSignalingOptions | Device.VSPSignalingOptions {
```
mhuynh5757, line 979: I like this idea of using smaller more modular methods to handle some of the options. But let's not mutate the options here and do that in the constructor. I see that you use this method to mutate options, but then also use it later to return an options object. I would recommend that in the constructor we do this:

```
const signalingOptions = this._resolveSignalingOptions(options);

this.options = { ...options, signalingOptions };
```

Something along these lines.

Then later, we don't ever need to use `this._resolveSignalingOptions()` again to check the signaling options. We just use `this._options`.

mhuynh5757 (changes requested):

lib/twilio/device.ts
```diff
@@ -1538,6 +1563,11 @@ class Device extends EventEmitter {
       this._destroyStream();
     }
 
+    if (this._options.signalingOptions?.useSignalingMethod === 'sip') {
+      this._log.info('Setting up SIP signaling');
+      // TODO(VBLOCKS-6371): Replace with SipSignalingAdapter
+    }
+
     this._log.info('Setting up VSP');
     const pstream = new (this._options.PStream || PStream)(
```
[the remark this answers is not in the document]
kpchoy: Will address in `TODO(VBLOCKS-6371)`

lib/twilio/device.ts
```diff
@@ -955,8 +956,32 @@ class Device extends EventEmitter {
...
+    }
+    if (!opts.sipServer) {
+      throw new InvalidArgumentError('`signalingOptions.sipServer` is required when `useSignalingMethod` is "sip"');
+    }
+    if (!opts.sipCredentials?.username || !opts.sipCredentials?.password) {
+      throw new InvalidArgumentError(
+        '`signalingOptions.sipCredentials` with `username` and `password` are required when `useSignalingMethod` is "sip"',
+      );
+    }
+    return opts;
```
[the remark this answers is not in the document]
kpchoy: done

tests/unit/device.ts
```diff
@@ -1595,7 +1667,7 @@ describe('Device', function() {
         });
 
         it('when it is an array', async () => {
-          await testWithOptions({ chunderw: ['foo', 'bar'] });
+          await testWithOptions({ signalingOptions: { useSignalingMethod: 'vsp', chunderw: ['foo', 'bar'] } });
           sinon.assert.calledOnceWithExactly(
```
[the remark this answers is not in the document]
kpchoy: done

lib/twilio/device.ts
```diff
@@ -2099,6 +2150,22 @@ namespace Device {
...
+     * @example SIP over WebSocket
+     * ```ts
+     * const sipDevice = new Device(token, {
+     *   signalingOptions: {
+     *     useSignalingMethod: 'sip',
+     *     sipServer: 'wss://sip.example.com',
+     *     sipCredentials: { username: 'foo', password: 'bar' },
+     *   },
+     * });
+     * ```
```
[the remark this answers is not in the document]
kpchoy: This will be implemented on this feature branch, so no need to address.

lib/twilio/device.ts
```diff
@@ -955,8 +956,37 @@ class Device extends EventEmitter {
...
    */
   private _getChunderws(): string[] | null {
-    return typeof this._options.chunderw === 'string' ? [this._options.chunderw]
-      : Array.isArray(this._options.chunderw) ? this._options.chunderw : null;
+    const vspOpts = this._options.signalingOptions?.useSignalingMethod === 'vsp'
+      ? this._options.signalingOptions as Device.VSPSignalingOptions
+      : undefined;
+    const chunderw = vspOpts?.chunderw ?? this._options.chunderw;
+    return typeof chunderw === 'string' ? [chunderw]
+      : Array.isArray(chunderw) ? chunderw : null;
```
mhuynh5757: This one is a little hard to read. Is it like this?

```
if (typeof chunderw === 'string') {
  return [chunderw];
}

if (Array.isArray(chunderw)) {
  // maybe we want to check if all the members are strings
  return chunderw;
}

return null;
```

lib/twilio/device.ts
```diff
@@ -955,8 +956,37 @@ class Device extends EventEmitter {
...
+
+  /**
+   * Resolves the active signaling options, defaulting to VSP if not specified.
+   * Throws {@link InvalidArgumentError} if SIP is selected but required fields are missing.
+   */
+  private _resolveSignalingOptions(): Device.SIPSignalingOptions | Device.VSPSignalingOptions {
+    const opts = this._options.signalingOptions;
+    if (!opts || opts.useSignalingMethod === 'vsp') {
+      return opts ?? { useSignalingMethod: 'vsp' };
+    }
```
mhuynh5757:
```suggestion
    if (!opts) {
      return { useSignalingMethod: 'vsp' };
    }

    if (opts.useSignalingMethod === 'vsp') {
      return opts;
    }
```

lib/twilio/device.ts
```diff
@@ -955,8 +956,37 @@ class Device extends EventEmitter {
...
+    if (opts.useSignalingMethod !== 'sip') {
+      throw new InvalidArgumentError(
+        `Unknown \`signalingOptions.useSignalingMethod\` value: "${(opts as any).useSignalingMethod}"`,
+      );
+    }
+    if (!opts.sipServer) {
```
mhuynh5757: Could check for `typeof === 'string'` for JS users. Same for below `username` and `password`.

mhuynh5757 (changes requested):

mhuynh5757 (approved):

PR #417, opened 2026-04-10
[VBLOCKS-6371] feat: sipSignalingAdapter connection + registration

kpchoy:
**Contributing to Twilio**

## Pull Request Details

https://twilio-engineering.atlassian.net/browse/VBLOCKS-6371

### Description

- Add sip.js@0.21.2 as a pinned dependency for SIP-over-WebSocket signaling
- Create SipSignalingAdapter (lib/twilio/signaling/sipsignalingadapter.ts) implementing the SignalingAdapter interface, covering WebSocket connection and SIP registration lifecycle via SIP.js UserAgent and Registerer
- Wire SipSignalingAdapter into Device._setupStream() so useSignalingMethod: 'sip' creates the SIP adapter instead of the PStream/VSP path
- Add optional region field to SIPSignalingOptions for edge/region metadata hints
- Call signaling methods (invite, answer, hangup, etc.) throw NotSupportedError as placeholders for VBLOCKS-6374
- Replace uglify-js with terser for minification to support ES6+ syntax from sip.js
- Exclude sip.js from rollup externals so it gets bundled inline (ESM → CJS) for the browserify dist build

### Before review
* [ ] Ready for review

### DIFF
--- a/lib/twilio/device.ts
+++ b/lib/twilio/device.ts
@@ -21,6 +21,7 @@ import { PreflightTest } from './preflight/preflight';
 import PStream from './pstream';
 import { PStreamSignalingAdapter } from './signaling/pstreamsignalingadapter';
 import { SignalingAdapter } from './signaling/signalingadapter';
+import { SipSignalingAdapter } from './signaling/sipsignalingadapter';
 import {
   createEventGatewayURI,
   createSignalingEndpointURL,
@@ -1583,19 +1584,24 @@ class Device extends EventEmitter {
 
     if (this._options.signalingOptions?.useSignalingMethod === 'sip') {
       this._log.info('Setting up SIP signaling');
-      // TODO(VBLOCKS-6371): Replace with SipSignalingAdapter
+      const sipOpts = this._options.signalingOptions as Device.SIPSignalingOptions;
+      this._stream = new SipSignalingAdapter({
+        sipServer: sipOpts.sipServer,
+        sipCredentials: sipOpts.sipCredentials,
+        region: sipOpts.region,
+      });
+    } else {
+      this._log.info('Setting up VSP');
+      const pstream = new (this._options.PStream || PStream)(
+        this.token,
+        this._chunderURIs,
+        {
+          backoffMaxMs: this._options.backoffMaxMs,
+          maxPreferredDurationMs: this._options.maxCallSignalingTimeoutMs,
+        },
+      );
+      this._stream = new PStreamSignalingAdapter(pstream);
     }
-
-    this._log.info('Setting up VSP');
-    const pstream = new (this._options.PStream || PStream)(
-      this.token,
-      this._chunderURIs,
-      {
-        backoffMaxMs: this._options.backoffMaxMs,
-        maxPreferredDurationMs: this._options.maxCallSignalingTimeoutMs,
-      },
-    );
-    this._stream = new PStreamSignalingAdapter(pstream);
 
     this._stream.addListener('close', this._onSignalingClose);
     this._stream.addListener('connected', this._onSignalingConnected);
@@ -1929,6 +1935,8 @@ namespace Device {
     sipServer: string;
     /** SIP digest authentication credentials. */
     sipCredentials: { username: string; password: string };
+    /** Optional region hint for edge/region metadata (e.g. `'ashburn'`). */
+    region?: string;
   }
 
   /**
--- a/lib/twilio/device.ts
+++ b/lib/twilio/device.ts
@@ -1573,6 +1573,35 @@ class Device extends EventEmitter {
   }
 
   /**
+   * Create the appropriate SignalingAdapter based on the configured signaling method.
+   */
+  private _createSignalingAdapter(): SignalingAdapter {
+    if (this._options.signalingOptions?.useSignalingMethod === 'sip') {
+      this._log.info('Setting up SIP signaling');
+      const sipOpts = this._options.signalingOptions as Device.SIPSignalingOptions;
+      const domain = new URL(sipOpts.sipServer).hostname;
+      return new SipSignalingAdapter({
+        sipDomain: domain,
+        sipUri: `sip:${sipOpts.sipCredentials.username}@${domain}`,
+        sipTransportServer: sipOpts.sipServer,
+        credentials: sipOpts.sipCredentials,
+        region: sipOpts.region || domain.split('.')[1],
+      });
+    }
+
+    this._log.info('Setting up VSP');
+    const pstream = new (this._options.PStream || PStream)(
+      this.token,
+      this._chunderURIs,
+      {
+        backoffMaxMs: this._options.backoffMaxMs,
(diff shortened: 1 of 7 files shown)


lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -0,0 +1,322 @@
...
+              this.emit('offline', this);
+            }
+            break;
+          case RegistererState.Terminated:
+            this._log.info('Registerer terminated');
+            break;
+        }
+      });
+
+      this._registerer.register().catch((error: Error) => {
```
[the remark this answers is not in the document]
kpchoy: In practice, register() is only called by Device after the connected event, and SIP.js does handle refresh internally once registered.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -0,0 +1,322 @@
...
+        this._log.error('SIP registration failed', error);
+        this.emit('error', {
+          error: {
+            code: 31201,
+            message: `SIP registration failed: ${error.message}`,
+          },
+        });
+      });
```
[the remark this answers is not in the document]
kpchoy: 👍

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -0,0 +1,322 @@
...
+    } else {
+      if (this._registerer) {
+        this._registerer.unregister().catch((error: Error) => {
+          this._log.error('SIP unregister failed', error);
+        });
+      }
```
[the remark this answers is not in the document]
kpchoy: 👍

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -0,0 +1,330 @@
...
+        this.emit('offline', this);
+      });
+    } else {
+      if (this._registerer) {
+        this._registerer.unregister().catch((error: Error) => {
+          this._log.error('SIP unregister failed', error);
+          this._status = 'offline';
+          this.emit('offline', this);
+        });
+      }
```
mhuynh5757, line 146: You can un-nest this by checking for the negative case first and then returning. That way you can assume the positive case and we don't have a huge chunk of logic in the `if-then` statement.

```
if (!isPresent) {
  ...
  return;
}
```

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -0,0 +1,330 @@
...
+    this._log.info('WebSocket disconnected', error?.message);
+
+    if (this._registerer) {
+      this._registerer.dispose().catch((err: Error) => {
+        this._log.warn('Error disposing registerer during disconnect', err);
+      });
+      this._registerer = null;
+    }
+
+    this.emit('transportClose');
```
mhuynh5757: Should this be gated by the `if` statement on line 325? Namely, should we only emit this when the status is not disconnected?

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -0,0 +1,330 @@
...
+  constructor(options: SipSignalingAdapterOptions) {
+    super();
+    this._options = options;
+    this._uri = options.sipServer;
+    this._region = options.region || this._extractRegionFromServer(options.sipServer);
+
+    const domain = extractDomain(options.sipServer);
+    const createUA = options.createUserAgent || ((config: any) => new UserAgent(config));
+
+    const sipUri = UserAgent.makeURI(`sip:${options.sipCredentials.username}@${domain}`);
```
mhuynh5757: I think all the logic with extracting and reconstructing the URIs and other options are a bit brittle and unnecessary. What do you think about putting those things in the `SipSignalingAdapterOptions` interface itself?

```
interface SipSignalingAdapterOptions {
  sipDomain: string;
  sipUri: string;
  sipTransportServer: string;
  credentials: { username: string; password: string }; // could flatten this if you want. purely a personal preference, i'm fine with or without flattening it
}
```

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -0,0 +1,330 @@
...
+  UserAgent,
+} from 'sip.js';
+import { NotSupportedError } from '../errors';
+import Log from '../log';
+import { SignalingAdapter, SignalingAdapterStatus } from './signalingadapter';
+
+/**
+ * @private
+ * Options for constructing a SipSignalingAdapter.
+ */
```
mhuynh5757, line 13: Personally I don't see a need to mark any of these as private since we're not exporting any of these and they shouldn't be picked up by the docs toolchain, but that might be worth fact-checking.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -0,0 +1,330 @@
...
+  sipCredentials: { username: string; password: string };
+  region?: string;
+  /**
+   * @private Factory override for testing.
+   */
+  createUserAgent?: (config: any) => any;
+  /**
+   * @private Factory override for testing.
+   */
+  createRegisterer?: (ua: any) => any;
```
mhuynh5757: Could we type this more narrowly?

```
createUserAgent?: (config: UserAgentOptions) => UserAgent;
```

I think that SIPjs exports these types for us, but if not, you can get away with some TS utility shenanigans.

```
createUserAgent?: (...args: ConstructorParameters<typeof UserAgent>) => UserAgent;
```

lib/twilio/device.ts
```diff
@@ -1583,20 +1584,25 @@ class Device extends EventEmitter {
...
+      this._log.info('Setting up VSP');
+      const pstream = new (this._options.PStream || PStream)(
+        this.token,
+        this._chunderURIs,
+        {
+          backoffMaxMs: this._options.backoffMaxMs,
+          maxPreferredDurationMs: this._options.maxCallSignalingTimeoutMs,
+        },
+      );
+      this._stream = new PStreamSignalingAdapter(pstream);
```
mhuynh5757: Maybe worth putting into a `setSignaling` private method or something. Also, PStream flow constructs the PStream and then passes it to the PStreamSignalingAdapter, where the SIP flow has the adapter construct the SIPjs objects itself. Not a huge deal, but is weirdly different. What do you think?

mhuynh5757 (changes requested):

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: this is cool I did not know about `ConstructorParameters` 👍

lib/twilio/device.ts
kpchoy: 👍  `_setupStream()` now just handles lifecycle (teardown, listener wiring, promise), and `_createSignalingAdapter()` handles the branching and construction.

luisrivas (approved):
Thanks, Kevin!

PR #420, opened 2026-04-16
[VBLOCKS-6374] feat: call signaling methods and events

kpchoy:
**Contributing to Twilio**

## Pull Request Details

https://twilio-engineering.atlassian.net/browse/VBLOCKS-6374

### Description

  - Implement call signaling methods (invite, answer, hangup, reject, reinvite, reconnect, dtmf, sendMessage) on SipSignalingAdapter, replacing NotSupportedError stubs
  - Track active sessions and pending inbound invitations via maps keyed by callSid
  - Map SIP.js session events to SignalingAdapter events (ringing, answer, hangup, cancel, ack, error) matching the payloads Call.ts and Device.ts expect
  - Clean up session state on hangup, destroy, and transport disconnect
  - SDP relay through SIP.js deferred to VBLOCKS-6372
  - Replace stub tests with full call signaling test coverage

### Before review
* [ ] Ready for review

### DIFF
--- a/lib/twilio/signaling/sipsignalingadapter.ts
+++ b/lib/twilio/signaling/sipsignalingadapter.ts
@@ -2,10 +2,12 @@ import { EventEmitter } from 'events';
 import {
   Invitation,
   Inviter,
+  InviterOptions,
   Registerer,
   RegistererState,
   Session,
   SessionState,
+  URI,
   UserAgent,
 } from 'sip.js';
 import Log from '../log';
@@ -31,7 +33,7 @@ export interface SipSignalingAdapterOptions {
   /**
    * Factory override for testing.
    */
-  createInviter?: (userAgent: UserAgent, targetURI: any, options?: any) => Inviter;
+  createInviter?: (userAgent: UserAgent, targetURI: URI, options?: InviterOptions) => Inviter;
 }
 
 /**
@@ -64,11 +66,11 @@ function createNoOpSDH(): any {
 export class SipSignalingAdapter extends EventEmitter implements SignalingAdapter {
   private _log: Log = new Log('SipSignalingAdapter');
   private _options: SipSignalingAdapterOptions;
-  private _outboundCallSids: Set<string> = new Set();
+  private _inboundSessions: Map<string, Invitation> = new Map();
+  private _outboundSessions: Map<string, Inviter> = new Map();
   private _pendingInvitations: Map<string, Invitation> = new Map();
   private _registerer: any | null = null;
   private _region: string | undefined;
-  private _sessions: Map<string, Session> = new Map();
   private _status: SignalingAdapterStatus = 'disconnected';
   private _token: string = '';
   private _uri: string;
@@ -207,16 +209,18 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
   destroy(): void {
     this._log.info('Destroying SipSignalingAdapter');
 
-    this._sessions.forEach((session) => {
+    const byeEstablished = (session: Session) => {
       if (session.state === SessionState.Established) {
         session.bye().catch((error: Error) => {
           this._log.warn('Error sending BYE during destroy', error);
         });
       }
-    });
-    this._sessions.clear();
+    };
+    this._inboundSessions.forEach(byeEstablished);
+    this._outboundSessions.forEach(byeEstablished);
+    this._inboundSessions.clear();
+    this._outboundSessions.clear();
     this._pendingInvitations.clear();
-    this._outboundCallSids.clear();
 
     if (this._registerer) {
       this._registerer.dispose().catch((error: Error) => {
@@ -259,7 +263,7 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
       return;
     }
     this._pendingInvitations.delete(callSid);
-    this._sessions.set(callSid, invitation);
+    this._inboundSessions.set(callSid, invitation);
     invitation.accept().catch((error: Error) => {
       this._log.error('Failed to accept invitation', error);
       this.emit('error', { error: { code: 31000, message: error.message }, callsid: callSid });
@@ -267,42 +271,19 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
   }
 
   hangup(callSid: string, _message?: string | null): void {
-    const invitation = this._pendingInvitations.get(callSid);
-    if (invitation) {
-      this._pendingInvitations.delete(callSid);
-      invitation.reject().catch((error: Error) => {
-        this._log.error('Failed to reject invitation during hangup', error);
-        this.emit('error', { error: { code: 31000, message: error.message }, callsid: callSid });
-      });
-      return;
+    if (this._pendingInvitations.has(callSid)) {
+      return this._hangupPendingInvitation(callSid);
     }
 
-    const session = this._sessions.get(callSid);
-    if (!session) {
-      this._log.warn('hangup: no session for callSid', callSid);
-      return;
+    if (this._outboundSessions.has(callSid)) {
+      return this._hangupOutboundSession(callSid);
     }
-    this._sessions.delete(callSid);
 
-    if (session.state === SessionState.Established) {
-      session.bye().catch((error: Error) => {
-        this._log.error('Failed to send BYE', error);
-        this.emit('error', { error: { code: 31000, message: error.message }, callsid: callSid });
-      });
-    } else if (session.state === SessionState.Initial || session.state === SessionState.Establishing) {
-      if (this._outboundCallSids.has(callSid)) {
-        (session as Inviter).cancel().catch((error: Error) => {
-          this._log.error('Failed to cancel outgoing call', error);
-          this.emit('error', { error: { code: 31000, message: error.message }, callsid: callSid });
-        });
-      } else {
-        (session as Invitation).reject().catch((error: Error) => {
-          this._log.error('Failed to reject invitation during hangup', error);
-          this.emit('error', { error: { code: 31000, message: error.message }, callsid: callSid });
-        });
-      }
+    if (this._inboundSessions.has(callSid)) {
+      return this._hangupInboundSession(callSid);
     }
-    this._outboundCallSids.delete(callSid);
+
+    this._log.warn('hangup: no session for callSid', callSid);
   }
 
   reject(callSid: string): void {
@@ -319,7 +300,7 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
   }
 
   reinvite(sdp: string, callSid: string): void {
-    const session = this._sessions.get(callSid);
+    const session = this._getSession(callSid);
     if (!session || session.state !== SessionState.Established) {
       this._log.warn('reinvite: no established session for callSid', callSid);
       return;
@@ -343,12 +324,12 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
   }
 
   dtmf(callSid: string, digits: string): void {
-    const session = this._sessions.get(callSid);
+    const session = this._getSession(callSid);
     if (!session || session.state !== SessionState.Established) {
       this._log.warn('dtmf: no established session for callSid', callSid);
       return;
     }
-    const sendDigit = async () => {
+    const sendDigits = async () => {
       for (const digit of digits) {
         try {
           await session.info({
@@ -365,7 +346,7 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
         }
       }
     };
-    sendDigit();
+    sendDigits();
   }
 
   sendMessage(
@@ -375,7 +356,7 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
     _messageType: string,
     voiceEventSid: string,
   ): void {
-    const session = this._sessions.get(callSid);
+    const session = this._getSession(callSid);
     if (!session || session.state !== SessionState.Established) {
       this._log.warn('sendMessage: no established session for callSid', callSid);
       return;
@@ -404,27 +385,81 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
   // Private helpers
   // ---------------------------------------------------------------------------
 
+  private _hangupPendingInvitation(callSid: string): void {
+    const invitation = this._pendingInvitations.get(callSid)!;
+    this._pendingInvitations.delete(callSid);
+    invitation.reject().catch((error: Error) => {
+      this._log.error('Failed to reject invitation during hangup', error);
+      this.emit('error', { error: { code: 31000, message: error.message }, callsid: callSid });
+    });
+  }
+
+  private _hangupOutboundSession(callSid: string): void {
+    const session = this._outboundSessions.get(callSid)!;
+    this._outboundSessions.delete(callSid);
+
+    if (session.state === SessionState.Established) {
+      session.bye().catch((error: Error) => {
+        this._log.error('Failed to send BYE', error);
+        this.emit('error', { error: { code: 31000, message: error.message }, callsid: callSid });
+      });
+    } else if (session.state === SessionState.Initial || session.state === SessionState.Establishing) {
+      session.cancel().catch((error: Error) => {
+        this._log.error('Failed to cancel outgoing call', error);
+        this.emit('error', { error: { code: 31000, message: error.message }, callsid: callSid });
+      });
+    }
+  }
+
+  private _hangupInboundSession(callSid: string): void {
+    const session = this._inboundSessions.get(callSid)!;
+    this._inboundSessions.delete(callSid);
+
+    if (session.state === SessionState.Established) {
+      session.bye().catch((error: Error) => {
+        this._log.error('Failed to send BYE', error);
+        this.emit('error', { error: { code: 31000, message: error.message }, callsid: callSid });
+      });
+    } else if (session.state === SessionState.Initial || session.state === SessionState.Establishing) {
+      session.reject().catch((error: Error) => {
+        this._log.error('Failed to reject invitation during hangup', error);
+        this.emit('error', { error: { code: 31000, message: error.message }, callsid: callSid });
+      });
+    }
+  }
+
+  private _getSession(callSid: string): Session | undefined {
+    return this._outboundSessions.get(callSid) || this._inboundSessions.get(callSid);
+  }
+
   private _handleIncomingInvite(invitation: Invitation): void {
-    const callSid = invitation.id;
+    const callSid = invitation.request.getHeader('X-Twilio-CallSid');
+    if (!callSid) {
+      this._log.error('Incoming INVITE missing X-Twilio-CallSid header');
+      invitation.reject().catch((error: Error) => {
+        this._log.error('Failed to reject invitation missing CallSid', error);
+      });
+      return;
+    }
     this._pendingInvitations.set(callSid, invitation);
 
-    invitation.delegate = {
-      ...invitation.delegate,
-      onCancel: () => {
-        this._pendingInvitations.delete(callSid);
-        this._sessions.delete(callSid);
-        this.emit('cancel', { callsid: callSid });
-      },
-      onBye: () => {
-        this._sessions.delete(callSid);
-        this.emit('hangup', { callsid: callSid });
-      },
+    if (!invitation.delegate) {
+      invitation.delegate = {};
+    }
+    invitation.delegate.onCancel = () => {
+      this._pendingInvitations.delete(callSid);
+      this._inboundSessions.delete(callSid);
+      this.emit('cancel', { callsid: callSid });
+    };
+    invitation.delegate.onBye = () => {
+      this._inboundSessions.delete(callSid);
+      this.emit('hangup', { callsid: callSid });
     };
 
     invitation.stateChange.addListener((state: SessionState) => {
       if (state === SessionState.Terminated) {
         this._pendingInvitations.delete(callSid);
-        this._sessions.delete(callSid);
+        this._inboundSessions.delete(callSid);
       }
     });
 
@@ -452,24 +487,21 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
     }
 
     const createInv = this._options.createInviter
-      || ((ua: UserAgent, uri: any, opts?: any) => new Inviter(ua, uri, opts));
+      || ((ua: UserAgent, uri: URI, opts?: InviterOptions) => new Inviter(ua, uri, opts));
     const inviter = createInv(this._userAgent, targetUri);
 
-    this._sessions.set(callSid, inviter);
-    this._outboundCallSids.add(callSid);
+    this._outboundSessions.set(callSid, inviter);
 
     inviter.delegate = {
       onBye: () => {
-        this._sessions.delete(callSid);
-        this._outboundCallSids.delete(callSid);
+        this._outboundSessions.delete(callSid);
         this.emit('hangup', { callsid: callSid });
       },
     };
 
     inviter.stateChange.addListener((state: SessionState) => {
       if (state === SessionState.Terminated) {
-        this._sessions.delete(callSid);
-        this._outboundCallSids.delete(callSid);
+        this._outboundSessions.delete(callSid);
       }
     });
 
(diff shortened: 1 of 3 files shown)


tests/unit/sipsignalingadapter.ts
```diff
@@ -301,48 +346,273 @@ describe('SipSignalingAdapter', () => {
...
+
+    it('should emit "answer" on requestDelegate.onAccept', (done) => {
+      const { adapter, inviterStub } = createAdapter();
+      adapter.on('answer', (payload: any) => {
+        assert.strictEqual(payload.callsid, 'call-1');
+        done();
+      });
+      adapter.invite('sdp', 'call-1', 'To=bob');
+      const rd = inviterStub._getRequestDelegate();
+      rd.onAccept({});
```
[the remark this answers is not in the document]
kpchoy: sdp will be implemented later PR

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -224,51 +245,258 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+    if (!targetUri) {
+      this._log.error('Failed to create target URI for invite');
+      this.emit('error', { error: { code: 31000, message: 'Invalid SIP target URI' }, callsid: callSid });
+      return;
+    }
+
+    const createInv = this._options.createInviter
+      || ((ua: UserAgent, uri: any, opts?: any) => new Inviter(ua, uri, opts));
+    const inviter = createInv(this._userAgent, targetUri);
+
```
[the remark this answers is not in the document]
kpchoy: sdp will be implemented later PR

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -224,51 +245,258 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+        onProgress: () => {
+          this.emit('ringing', { callsid: callSid });
+        },
+        onAccept: () => {
+          const payload: Record<string, any> = { callsid: callSid, sdp: '' };
+          if (reconnectToken) {
+            payload.reconnect = reconnectToken;
+          }
+          this.emit('answer', payload);
+        },
```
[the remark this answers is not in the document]
kpchoy: sdp will be implemented later PR

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -224,51 +245,258 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+    }
+    session.invite({
+      requestDelegate: {
+        onAccept: () => {
+          this.emit('answer', { callsid: callSid, sdp: '' });
+        },
+        onReject: (response: any) => {
+          this._log.warn('re-INVITE rejected', response?.message?.statusCode);
+        },
+      },
```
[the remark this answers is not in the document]
kpchoy: sdp will be implemented later PR

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -224,51 +245,258 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+    });
+
+    this.emit('invite', {
+      callsid: callSid,
+      sdp: invitation.body || '',
+      parameters: {
+        CallSid: callSid,
+        From: invitation.remoteIdentity.uri.toString(),
+      },
+    });
```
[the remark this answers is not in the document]
kpchoy: In practice, SIP INVITEs always carry an SDP body. More importantly, once VBLOCKS-6372 lands, SDP will flow through the SDH rather than being pulled from invitation.body directly. Adding a guard now would be premature defensive code that gets rewritten in the next ticket.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -224,51 +245,258 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
-    throw new NotSupportedError('SIP call signaling is not yet implemented (VBLOCKS-6374)');
+  answer(sdp: string, callSid: string): void {
+    const invitation = this._pendingInvitations.get(callSid);
+    if (!invitation) {
+      this._log.warn('answer: no pending invitation for callSid', callSid);
+      return;
+    }
+    this._pendingInvitations.delete(callSid);
+    this._sessions.set(callSid, invitation);
+    invitation.accept().catch((error: Error) => {
```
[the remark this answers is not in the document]
kpchoy: The invitation already has a stateChange listener from _handleIncomingInvite that cleans up on SessionState.Terminated. If accept() fails, SIP.js will transition the session to Terminated, triggering that cleanup. So the stale entry concern is already handled by the existing listener.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -224,51 +245,258 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+    callSid: string,
+    content: string,
+    contentType: string | undefined,
     _messageType: string,
-    _voiceEventSid: string,
+    voiceEventSid: string,
   ): void {
-    throw new NotSupportedError('SIP call signaling is not yet implemented (VBLOCKS-6374)');
+    const session = this._sessions.get(callSid);
+    if (!session) {
```
charliesantos: should we check for session state as well making sure it's established?

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -224,51 +245,258 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+      });
+    });
   }
 
   // ---------------------------------------------------------------------------
   // Private helpers
   // ---------------------------------------------------------------------------
 
+  private _handleIncomingInvite(invitation: Invitation): void {
+    const callSid = invitation.id;
```
charliesantos: I thought Invitation.id is the SIP Call-ID, not the Twilio CallSid?

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -224,51 +245,258 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+      session.bye().catch((error: Error) => {
+        this._log.warn('Failed to send BYE', error);
+      });
+    } else if (session.state === SessionState.Initial || session.state === SessionState.Establishing) {
+      if (this._outboundCallSids.has(callSid)) {
+        (session as Inviter).cancel().catch((error: Error) => {
+          this._log.warn('Failed to cancel outgoing call', error);
+        });
+      } else {
+        (session as unknown as Invitation).reject().catch((error: Error) => {
```
charliesantos: double casts are usually a smell. why do we need it here?

charliesantos (changes requested):

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -25,12 +28,18 @@ export interface SipSignalingAdapterOptions {
    * Factory override for testing.
    */
   createRegisterer?: (userAgent: UserAgent) => Registerer;
+  /**
+   * Factory override for testing.
+   */
+  createInviter?: (userAgent: UserAgent, targetURI: any, options?: any) => Inviter;
```
mhuynh5757: Can we more explicitly type the parameters in this method?

charliesantos (changes requested):
Also review from Claude. Let me know your thoughts @kpchoy 

Design / Completeness                                                                                                                     
                                                                                                                                            
  1. SDP and params silently dropped — invite(), answer(), reinvite(), reconnect() all accept sdp/params but never pass them through. The PR
   description says this is deferred to VBLOCKS-6372, but the parameters are silently consumed with no TODO or assertion. A future dev might
   assume they're wired up. Consider adding a brief inline comment or at minimum a _log.debug noting SDP relay is deferred.
  2. params not used in _sendInvite() — The params string (e.g., "To=bob") is accepted but never incorporated into the SIP INVITE headers or
   URI. Is this intentional for this ticket?                                                                                                
  3. _messageType silently ignored in sendMessage() — The underscore prefix signals intent, but there's no indication if this is permanently
   unused or deferred.                                                                                                                      
  4. Inconsistent error handling in hangup() — When bye(), cancel(), or reject() fail, the errors are only logged. In contrast, answer() and
   reject() (the standalone method) emit 'error' events. Consider emitting errors consistently, especially for bye() failures which the     
  caller may need to know about.
                                                                                                                                            
  Tests                     

  5. Missing failure-path coverage — No tests for:                                                                                          
    - accept() rejecting in answer() (should emit error)
    - reject() rejecting in reject() (should emit error)                                                                                    
    - reinvite() requestDelegate.onReject callback                                                                                          
    - onBye delegate firing after an answered inbound call                                                                                  
  6. Fragile stub mutation in sendMessage error test (sipsignalingadapter.ts:~L595) — The test creates a failInviterStub, calls             
  adapter.invite() to register it in _sessions, then mutates .state directly. This couples the test to internal map storage. Consider       
  exposing a helper or using _simulateState.                                                                                                
                                                                                                                                            
  Nits                      

  - createInviter factory uses any for targetURI and options — could use import type { URI } from SIP.js for better type safety.            
  - The sendDigit async function in dtmf() is fire-and-forget — its returned promise is never handled. If a digit fails, subsequent digits
  still attempt. This is probably fine for DTMF but worth a brief comment.                                                                  
                            
  Summary                                                                                                                                   
                            
  The core implementation is well-structured and the state machine logic is correct. The main concerns are: (1) session state check missing 
  in sendMessage, (2) confirming invitation.id actually maps to Twilio CallSid, and (3) a few gaps in error-path test coverage. Everything
  else is minor.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -224,51 +245,258 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
-  answer(_sdp: string, _callSid: string): void {
-    throw new NotSupportedError('SIP call signaling is not yet implemented (VBLOCKS-6374)');
+  answer(sdp: string, callSid: string): void {
+    const invitation = this._pendingInvitations.get(callSid);
+    if (!invitation) {
+      this._log.warn('answer: no pending invitation for callSid', callSid);
+      return;
+    }
+    this._pendingInvitations.delete(callSid);
+    this._sessions.set(callSid, invitation);
```
mhuynh5757: Are `Invitations` and `Sessions` the same object? `_pendingInvitations` is a map that contains `Invitations` but then `_sessions` is a map containing `Sessions` and it's just being taken out of one map here and put into the other.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -224,51 +245,258 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+    invitation.accept().catch((error: Error) => {
+      this._log.error('Failed to accept invitation', error);
+      this.emit('error', { error: { code: 31000, message: error.message }, callsid: callSid });
+    });
   }
 
-  hangup(_callSid: string, _message?: string | null): void {
-    throw new NotSupportedError('SIP call signaling is not yet implemented (VBLOCKS-6374)');
+  hangup(callSid: string, _message?: string | null): void {
```
mhuynh5757, line 273: Let me know how you feel about this, not a direct recommendation, curious how you feel about reorganizing methods like this:

The hangup method seems to handle two branches, rejecting a pending invite and hanging up an ongoing call. What if the method looks like this:

```
hangup(callSid: string, _message?: string | null) : void {
  if (this._pendingInvitations.has(callSid)) {
    return this._hangupPendingInvite();
  }

  if (this._sessions.has(callSid)) {
    return this._hangupOngoingCall();
  }

  this._log.warn(...);
  // maybe emit error or throw?
  return;
}

Comment (mhuynh5757) [reply lib/twilio/signaling/sipsignalingadapter.ts:273]: I think having a method for each branch would help our ability to maintain down the line. Right now when we dig through the existing code it's hard to tell where one codepath ends and another begins. What do you think?

File: lib/twilio/signaling/sipsignalingadapter.ts [hunk 13]
```diff
@@ -224,51 +245,258 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+      });
+    } else if (session.state === SessionState.Initial || session.state === SessionState.Establishing) {
+      if (this._outboundCallSids.has(callSid)) {
+        (session as Inviter).cancel().catch((error: Error) => {
+          this._log.warn('Failed to cancel outgoing call', error);
+        });
+      } else {
+        (session as unknown as Invitation).reject().catch((error: Error) => {
+          this._log.warn('Failed to reject invitation during hangup', error);
+        });
```
mhuynh5757: The type-casting here is giving a weird vibe. Is there no better way to distinguish these session objects?

mhuynh5757: Just an idea, maybe a discriminated union? `{ type: 'invitation', session: Invitation } | { type: 'inviter', session: Inviter }`. Or maybe a better idea is that we'll need multiple session maps for each type of session?

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -224,51 +245,258 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
 
-  dtmf(_callSid: string, _digits: string): void {
-    throw new NotSupportedError('SIP call signaling is not yet implemented (VBLOCKS-6374)');
+  dtmf(callSid: string, digits: string): void {
+    const session = this._sessions.get(callSid);
+    if (!session || session.state !== SessionState.Established) {
+      this._log.warn('dtmf: no established session for callSid', callSid);
+      return;
+    }
+    const sendDigit = async () => {
```
mhuynh5757: Nit: `sendDigits` plural instead of singular.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -224,51 +245,258 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+      onCancel: () => {
+        this._pendingInvitations.delete(callSid);
+        this._sessions.delete(callSid);
+        this.emit('cancel', { callsid: callSid });
+      },
+      onBye: () => {
+        this._sessions.delete(callSid);
+        this.emit('hangup', { callsid: callSid });
+      },
+    };
```
mhuynh5757: Nit: can we separate this?

```
invitation.delegate.onCancel = () => {
...
};

invitation.delegate.onBye = () => {
...
};
```

That way we don't reassign the whole delegate object and make an unnecessary copy. I don't know if this will actually affect anything under the hood, maybe a unit-test injection or something, so I'm labeling this suggestion as a nitpick.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -224,51 +245,258 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+
+    const targetUri = UserAgent.makeURI(`sip:${this._options.sipDomain}`);
+    if (!targetUri) {
+      this._log.error('Failed to create target URI for invite');
+      this.emit('error', { error: { code: 31000, message: 'Invalid SIP target URI' }, callsid: callSid });
+      return;
+    }
+
+    const createInv = this._options.createInviter
+      || ((ua: UserAgent, uri: any, opts?: any) => new Inviter(ua, uri, opts));
```
mhuynh5757: Same as other comment about explicitly typing these parameters.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -224,51 +245,258 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+    this._sessions.set(callSid, inviter);
+    this._outboundCallSids.add(callSid);
+
+    inviter.delegate = {
+      onBye: () => {
+        this._sessions.delete(callSid);
+        this._outboundCallSids.delete(callSid);
+        this.emit('hangup', { callsid: callSid });
+      },
+    };
```
mhuynh5757, line 524: Same as other comment

```
inviter.delegate.onBye = () => {
...
}
```

Unless there's no delegate object here initially?

mhuynh5757 (changes requested):

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: Yes the invitation.id is the SIP Call-ID header, not the Twilio CallSid. For incoming SIP INVITEs, the Twilio CallSid would need to come from the SIP message — likely via a custom header (e.g., X-Twilio-CallSid) accessible through invitation.request.getHeader(). Do you know what this would look like from the Backend?

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: Good catch 👍

kpchoy:
@charliesantos I addressed #4, #5, #6. The other comments will be handled in later tickets

  #4 — Consistent error handling in hangup(): All three failure paths (bye(), cancel(), reject()) now emit 'error' events with { code: 31000 }, matching answer() and reject().

  #5 — Missing failure-path tests (5 new tests):
  - answer(): emit error if accept() fails
  - reject(): emit error if reject() fails
  - reinvite(): onAccept emits 'answer'
  - reinvite(): onReject warns but doesn't throw
  - Incoming call: onBye fires correctly after answering an inbound call

  #6 — Fragile stub mutation: Changed (failInviterStub as any).state = 'Established' to failInviterStub._simulateState('Established').

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: actually just asked Andreas 👍

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: Yes, `Invitation` extends `Session` in SIP.js — it's the same object instance. The two maps represent different lifecycle stages: `_pendingInvitations` holds inbound calls that haven't been answered yet, and `_sessions` holds active calls (both inbound after `answer()` and outbound after `invite()`). When `answer()` is called, the invitation moves from one map to the other. It's stored as `Session` in `_sessions` because that's the common interface shared with outbound `Inviter` instances.

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: I think it is a good idea 👍

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: Lets do multiple session maps for each type of session

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -224,51 +249,288 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
   }
 
   sendMessage(
-    _callSid: string,
-    _content: string,
-    _contentType: string | undefined,
+    callSid: string,
+    content: string,
+    contentType: string | undefined,
     _messageType: string,
```
charliesantos: I see this is unused. Should we remove? If for future, can we comment?

charliesantos (changes requested):

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: `_messageType` is required by the `SignalingAdapter` interface so we can't remove it from the signature. The `_` prefix indicates it's intentionally unused. The SIP MESSAGE body doesn't have a direct equivalent for this field — it's specific to the PStream protocol.

kpchoy: added comment

charliesantos: Ah got it. I have a concern with this but feel free to address it here or in the future.
If more PStream-specific fields leak into SignalingAdapter, SIP implementations will accumulate more dead params. Another approach would be to create a options object so that each adapter destructures only what it needs and adding new fields doesn't change the signature. For example

```ts
                                                                                                                                            
  interface SendMessageOptions {                                                                                                            
    content: string;                                                                                                                        
    contentType?: string;                                                                                                                   
    messageType: string;                                                                                                                    
    voiceEventSid: string;                                                                                                                  
  }                                                                                                                                         
                                                                                                                                            
  sendMessage(callSid: string, options: SendMessageOptions): void {
    const { content, contentType, voiceEventSid } = options;
  }
```

kpchoy: I like the idea. Seems like it will be a bigger change so will do this in a different ticket.

@mhuynh5757 since you wrote the adapter wdyt? I can create a ticket if we want to move forward with this idea.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -400,27 +385,81 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
   // Private helpers
   // ---------------------------------------------------------------------------
 
+  private _hangupPendingInvitation(callSid: string): void {
+    const invitation = this._pendingInvitations.get(callSid)!;
```
mhuynh5757: I know that we only use this method within the conditional, making this `!` relatively safe, but let's see if we can avoid it if we can, just in case we decide to invoke this method from elsewhere.

What if we change the signature to be stateless?

```
private _hangupPendingInvitation(invitation: Invitation): void {
  ...
}
```

When we use in the above code, since we already checked for the nonnullable case, TS should narrow this and it should work out.

```
const pendingInvitation = this._pendingInvitations.get(callSid);
if (pendingInvitation) {
  return this._hangupPendingInvitation(pendingInvitation);
}
```

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -400,27 +385,81 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+      session.bye().catch((error: Error) => {
+        this._log.error('Failed to send BYE', error);
+        this.emit('error', { error: { code: 31000, message: error.message }, callsid: callSid });
+      });
+    } else if (session.state === SessionState.Initial || session.state === SessionState.Establishing) {
+      session.cancel().catch((error: Error) => {
+        this._log.error('Failed to cancel outgoing call', error);
+        this.emit('error', { error: { code: 31000, message: error.message }, callsid: callSid });
+      });
+    }
```
mhuynh5757: I see that there are other SessionStates that are not covered here. Paraphrasing from the SIPjs code:

```
enum SessionState {
  'Initial',
  'Establishing',
  'Established',
  'Terminating',
  'Terminated',
}
```

Can we leverage a switch-case and have TS do some build-time exhaustivity checks?

```
switch (session.state) {
  case 'Established': {
    ...
    return;
  }
  case 'Initial':
  case 'Establishing: {
    return;
  }
  case 'Terminating':
  case 'Terminated': {
    // no-op, or maybe silently warn that we have a weird state?
  }
  default: {
    const _state: never = session.state; // if we don't explicitly handle all of the states, then this type is not properly narrowed down to "never" and typescript will throw an error at build time, guaranteeing we handle the states
    // warn or throw an error about a funky state
  }
}
```

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -400,27 +385,81 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+      session.bye().catch((error: Error) => {
+        this._log.error('Failed to send BYE', error);
+        this.emit('error', { error: { code: 31000, message: error.message }, callsid: callSid });
+      });
+    } else if (session.state === SessionState.Initial || session.state === SessionState.Establishing) {
+      session.reject().catch((error: Error) => {
+        this._log.error('Failed to reject invitation during hangup', error);
+        this.emit('error', { error: { code: 31000, message: error.message }, callsid: callSid });
+      });
+    }
```
mhuynh5757: Same as above comments.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -400,27 +385,81 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+  }
+
+  private _getSession(callSid: string): Session | undefined {
+    return this._outboundSessions.get(callSid) || this._inboundSessions.get(callSid);
```
mhuynh5757, line 456: Really like this helper method, very nice.

mhuynh5757 (changes requested):

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: https://twilio-engineering.atlassian.net/browse/VBLOCKS-6500

mhuynh5757 (approved):

PR #425, opened 2026-04-21
Fetch OIDC token

charliesantos:
**Contributing to Twilio**

- [ ] I acknowledge that all my contributions will be made under the project's license.

## Pull Request Details

A description of what this PR does including issue links (JIRA, github, etc) if applicable.

### Description

A description of what this PR does.

### Before review
* [ ] Ready for review

### DIFF
--- a/.release.json
+++ b/.release.json
@@ -53,7 +53,6 @@
     "publish": {
       "commands": [
         "git checkout ${RELEASE_VERSION}",
-        "echo \"//registry.npmjs.org/:_authToken=${NPM_AUTH_TOKEN}\" >~/.npmrc",
         "npm publish"
       ]
     }
--- a/.release.json
+++ b/.release.json
@@ -53,7 +53,7 @@
     "publish": {
       "commands": [
         "git checkout ${RELEASE_VERSION}",
-        "export NPM_ID_TOKEN=$(circleci run oidc get --claims '{\"aud\":\"npm:registry.npmjs.org\"}') && npm publish"
+        "NPM_ID_TOKEN=$(circleci run oidc get --claims '{\"aud\":\"npm:registry.npmjs.org\"}') npm publish"
       ]
     }
   }
--- a/.release.json
+++ b/.release.json
@@ -53,7 +53,7 @@
     "publish": {
       "commands": [
         "git checkout ${RELEASE_VERSION}",
-        "npm publish"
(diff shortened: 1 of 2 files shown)


.release.json
```diff
@@ -53,7 +53,6 @@
     "publish": {
       "commands": [
         "git checkout ${RELEASE_VERSION}",
-        "echo \"//registry.npmjs.org/:_authToken=${NPM_AUTH_TOKEN}\" >~/.npmrc",
         "npm publish"
```
mhuynh5757: What do you think about having the `NPM_ID_TOKEN` only set _just_ for this command. Like so:
```
"NPM_ID_TOKEN=$(circleci run oidc get --claims '{"aud": "npm:registry.npmjs.org"}') npm publish"
```
And we don't do it at all in the `.circleci/config.yml` file?

mhuynh5757: Feels like less potential for leaks, considering the way we do it in the `config.yml` puts it in the session-persistent `BASH_ENV`.

mhuynh5757 (changes requested):

.release.json
```diff
@@ -53,8 +53,7 @@
     "publish": {
       "commands": [
         "git checkout ${RELEASE_VERSION}",
-        "echo \"//registry.npmjs.org/:_authToken=${NPM_AUTH_TOKEN}\" >~/.npmrc",
-        "npm publish"
+        "export NPM_ID_TOKEN=$(circleci run oidc get --claims '{\"aud\":\"npm:registry.npmjs.org\"}') && npm publish"
```
mhuynh5757:
```suggestion
        "NPM_ID_TOKEN=$(circleci run oidc get --claims '{\"aud\":\"npm:registry.npmjs.org\"}') npm publish"
```

Sorry I know this looks really nitpicky but I think this is better long term since if we use `export ... &&` we're keeping the `NPM_ID_TOKEN` env var longer lived than it should be. If any commands happen in the box after `npm publish` the secret will be present. Granted, the token is single-use (I think), so this is probably not worth addressing. But I wanted to bring it up anyway.

kpchoy (approved):

mhuynh5757 (changes requested):

.release.json
charliesantos: Fair point, tighter scope is strictly better, even if practically the risk is low here (the entry runs in its own subshell, and OIDC tokens expire in an hour). I'll update it.

mhuynh5757 (approved):

PR #427, opened 2026-04-22
[VBLOCKS-6500] refactor: callsignaling methods to config objects

kpchoy:
**Contributing to Twilio**

## Pull Request Details

https://twilio-engineering.atlassian.net/browse/VBLOCKS-6500

### Description

  - Refactor SignalingAdapter call-signaling methods from positional params to config objects (InviteConfig, AnswerConfig, HangupConfig, ReinviteConfig, ReconnectConfig, DtmfConfig, SendMessageConfig)
  - callSid remains as the first positional param; reject() unchanged (single param)
  - PStreamSignalingAdapter destructures each config and forwards positionally to PStream's existing API
  - SipSignalingAdapter destructures only the fields it needs; introduces SipSendMessageConfig to narrow the type (excludes PStream-specific messageType)
  - Call sites in call.ts build typed consts (e.g. const inviteConfig: InviteConfig = { ... }) before passing to adapter methods
  - Internal-only interface — no public API changes

### Before review
* [ ] Ready for review

### DIFF
--- a/lib/twilio/call.ts
+++ b/lib/twilio/call.ts
@@ -2,7 +2,16 @@ import { EventEmitter } from 'events';
 import Backoff from './backoff';
 import Device from './device';
 import DialtonePlayer from './dialtonePlayer';
-import { SignalingAdapter } from './signaling/signalingadapter';
+import {
+  AnswerConfig,
+  DtmfConfig,
+  HangupConfig,
+  InviteConfig,
+  ReconnectConfig,
+  ReinviteConfig,
+  SendMessageConfig,
+  SignalingAdapter,
+} from './signaling/signalingadapter';
 import {
   GeneralErrors,
   getPreciseSignalingErrorByCode,
@@ -373,7 +382,8 @@ class Call extends EventEmitter {
 
         this._signalingAdapter.on('answer', onAnswerOrRinging);
         this._signalingAdapter.on('hangup', onHangup);
-        this._signalingAdapter.reinvite(offerSdp, this.parameters.CallSid);
+        const reinviteConfig: ReinviteConfig = { sdp: offerSdp };
+        this._signalingAdapter.reinvite(this.parameters.CallSid, reinviteConfig);
       });
     });
 
@@ -682,7 +692,8 @@ class Call extends EventEmitter {
         this._mediaHandler.answerIncomingCall(this.parameters.CallSid,
           this._options.offerSdp, rtcConfiguration,
           (answerSdp: string) => {
-            this._signalingAdapter.answer(answerSdp, this.parameters.CallSid);
+            const answerConfig: AnswerConfig = { sdp: answerSdp };
+            this._signalingAdapter.answer(this.parameters.CallSid, answerConfig);
           },
           onAnswer);
       } else {
@@ -708,9 +719,11 @@ class Call extends EventEmitter {
           this._signalingAdapter.on('ringing', onPstreamAnswerOrRinging);
 
           if (this._signalingReconnectToken) {
-            this._signalingAdapter.reconnect(offerSdp, outgoingCallSid, this._signalingReconnectToken);
+            const reconnectConfig: ReconnectConfig = { sdp: offerSdp, reconnectToken: this._signalingReconnectToken };
+            this._signalingAdapter.reconnect(outgoingCallSid, reconnectConfig);
           } else {
-            this._signalingAdapter.invite(offerSdp, outgoingCallSid, params);
+            const inviteConfig: InviteConfig = { sdp: offerSdp, params };
+            this._signalingAdapter.invite(outgoingCallSid, inviteConfig);
           }
 
           // NOTE(VBLOCKS-6417): in the original implementation, the RTC DTLS
@@ -947,7 +960,8 @@ class Call extends EventEmitter {
     this._log.info('Sending digits over PStream');
 
     if (this._signalingAdapter !== null && this._signalingAdapter.status !== 'disconnected') {
-      this._signalingAdapter.dtmf(this.parameters.CallSid, digits);
+      const dtmfConfig: DtmfConfig = { digits };
+      this._signalingAdapter.dtmf(this.parameters.CallSid, dtmfConfig);
     } else {
       const error = new GeneralErrors.ConnectionError('Could not send DTMF: Signaling channel is disconnected');
       this._log.debug('#error', error);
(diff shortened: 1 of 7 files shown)


lib/twilio/signaling/pstreamsignalingadapter.ts
```diff
@@ -51,41 +61,41 @@ export class PStreamSignalingAdapter extends EventEmitter implements SignalingAd
...
-  answer(sdp: string, callSid: string): void {
+  answer(callSid: string, config: AnswerConfig): void {
+    const { sdp } = config;
     this._pstream.answer(sdp, callSid);
   }
 
-  hangup(callSid: string, message?: string | null): void {
-    this._pstream.hangup(callSid, message);
+  hangup(callSid: string, config?: HangupConfig): void {
+    this._pstream.hangup(callSid, config?.message);
```
mhuynh5757: What do you think about this instead, so it fits the rest of the methods:
```
hangup(callSid: string; config: HangupConfig = {}): void {
  const { message } = config;
  this._pstream.hangup(callSid, message);
}
```

I'm personally not a fan of optional (and consequently default parameter values) but since that one is more of a personal choice I'm not going to recommend it.

lib/twilio/signaling/signalingadapter.ts
```diff
@@ -2,6 +2,39 @@ import { EventEmitter } from 'events';
...
+  sdp: string;
+  params: string;
+}
+
+export interface AnswerConfig {
+  sdp: string;
+}
+
+export interface HangupConfig {
+  message?: string | null;
```
mhuynh5757: Hmm, double-nullish values seems kind of strange. Any way we can reduce this to `message?: string` or is `null` required?

lib/twilio/signaling/signalingadapter.ts
```diff
@@ -2,6 +2,39 @@ import { EventEmitter } from 'events';
...
+  reconnectToken: string;
+}
+
+export interface DtmfConfig {
+  digits: string;
+}
+
+export interface SendMessageConfig {
+  content: string;
+  contentType: string | undefined;
```
mhuynh5757: Do we want to make this `contentType?: string`?

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -11,7 +11,22 @@ import {
...
+  ReinviteConfig,
+  SignalingAdapter,
+  SignalingAdapterStatus,
+} from './signalingadapter';
+
+interface SipSendMessageConfig {
+  content: string;
+  contentType: string | undefined;
+  voiceEventSid: string;
+}
```
mhuynh5757: Because we absolutely want this to be a derivative of the larger SendMessageConfig, I think we should either use Omit or Pick to reference the larger config type.

```
type SipSendMessageConfig = Pick<SendMessageConfig, 'content' | 'contentType' | 'voiceEventSid'>;
```

That way if we ever need to change `'contentType'` to a stricter typing, for example `'base64' | 'json' | ...` then this change propagates down and we don't need to worry about forgetting it.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -252,11 +267,12 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
   // ---------------------------------------------------------------------------
 
-  invite(sdp: string, callSid: string, params: string): void {
+  invite(callSid: string, config: InviteConfig): void {
+    const { sdp, params } = config;
     this._sendInvite(sdp, callSid, params);
   }
 
-  answer(sdp: string, callSid: string): void {
+  answer(callSid: string, config: AnswerConfig): void {
```
mhuynh5757, line 271: It looks like we never used `sdp` in the prior implementation and now we're not using `config`? Is that similar to `hangup` and why the config in `hangup` is set to be optional?

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -302,7 +318,7 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
     });
   }
 
-  reinvite(sdp: string, callSid: string): void {
+  reinvite(callSid: string, config: ReinviteConfig): void {
```
mhuynh5757, line 317: Same question as above.

lib/twilio/call.ts
```diff
@@ -997,7 +1011,8 @@ class Call extends EventEmitter {
 
     const voiceEventSid = this._voiceEventSidGenerator();
     this._messages.set(voiceEventSid, { content, contentType, messageType, voiceEventSid });
-    this._signalingAdapter.sendMessage(callSid!, content, contentType, messageType, voiceEventSid);
+    const sendMessageConfig: SendMessageConfig = { content, contentType, messageType, voiceEventSid };
+    this._signalingAdapter.sendMessage(callSid!, sendMessageConfig);
```
mhuynh5757, line 1015: I wonder if we should have added a proper null-check here. Maybe not worth worrying about in this PR but something to toss on the backlog? "Check for any `!` operators and put in proper null-checks"? WDYT?

mhuynh5757 (changes requested):

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: In the next ticket we will need SDP, it is a placeholder

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: In the next ticket we will need SDP, it is a placeholder

lib/twilio/call.ts
kpchoy: I like it 👍 : https://twilio-engineering.atlassian.net/browse/VBLOCKS-6507

lib/twilio/signaling/pstreamsignalingadapter.ts
```diff
@@ -51,41 +61,42 @@ export class PStreamSignalingAdapter extends EventEmitter implements SignalingAd
     this._pstream.updateURIs(uris);
   }
 
-  invite(sdp: string, callSid: string, params: string): void {
+  invite(callSid: string, config: InviteConfig): void {
```
charliesantos: You can destructure directly in the parameter list so you don't have to destructure inside the function. let's use this pattern throughout this PR to prevent duplication.

```ts
invite(callSid: string, { sdp, params }: InviteConfig): void {
```

lib/twilio/signaling/signalingadapter.ts
```diff
@@ -2,6 +2,39 @@ import { EventEmitter } from 'events';
...
+export interface DtmfConfig {
+  digits: string;
+}
+
+export interface SendMessageConfig {
+  content: string;
+  contentType?: string;
+  messageType: string;
+  voiceEventSid: string;
+}
```
charliesantos, line 36: You can DRY this up with inheritance, but these types are simple enough that extra abstraction may hurt readability more than it helps. Worth revisiting if/when the configs grow more complex.

charliesantos (changes requested):

charliesantos (approved):

mhuynh5757 (approved):

PR #428, opened 2026-04-24
[VBLOCKS-6507] feat: null checks call

kpchoy:
**Contributing to Twilio**

## Pull Request Details

https://twilio-engineering.atlassian.net/browse/VBLOCKS-6507

### Description

Add null checks to call

### Before review
* [ ] Ready for review

### DIFF
--- a/lib/twilio/call.ts
+++ b/lib/twilio/call.ts
@@ -382,8 +382,13 @@ class Call extends EventEmitter {
 
         this._signalingAdapter.on('answer', onAnswerOrRinging);
         this._signalingAdapter.on('hangup', onHangup);
+        const callSid = this.parameters.CallSid;
+        if (!callSid) {
+          this._log.warn('Cannot reinvite: CallSid is not set');
+          return;
+        }
         const reinviteConfig: ReinviteConfig = { sdp: offerSdp };
-        this._signalingAdapter.reinvite(this.parameters.CallSid, reinviteConfig);
+        this._signalingAdapter.reinvite(callSid, reinviteConfig);
       });
     });
 
@@ -665,11 +670,13 @@ class Call extends EventEmitter {
         this._publisher.info('connection', eventName, null, this);
 
         // Report the preferred codec and params as they appear in the SDP
-        const { codecName, codecParams } = getPreferredCodecInfo(this._mediaHandler.version.getSDP());
-        this._publisher.info('settings', 'codec', {
-          codec_params: codecParams,
-          selected_codec: codecName,
-        }, this);
+        if (this._mediaHandler.version) {
+          const { codecName, codecParams } = getPreferredCodecInfo(this._mediaHandler.version.getSDP());
+          this._publisher.info('settings', 'codec', {
+            codec_params: codecParams,
+            selected_codec: codecName,
+          }, this);
+        }
 
         // Enable RTC monitoring
         this._monitor.enable(pc);
@@ -687,13 +694,18 @@ class Call extends EventEmitter {
       this._signalingAdapter.addListener('hangup', this._onHangup);
 
       if (this._direction === Call.CallDirection.Incoming) {
+        const callSid = this.parameters.CallSid;
+        if (!callSid) {
+          this._log.error('Cannot answer incoming call: CallSid is not set');
+          return;
+        }
         this._isAnswered = true;
         this._signalingAdapter.on('answer', this._onAnswer);
-        this._mediaHandler.answerIncomingCall(this.parameters.CallSid,
+        this._mediaHandler.answerIncomingCall(callSid,
           this._options.offerSdp, rtcConfiguration,
           (answerSdp: string) => {
             const answerConfig: AnswerConfig = { sdp: answerSdp };
-            this._signalingAdapter.answer(this.parameters.CallSid, answerConfig);
+            this._signalingAdapter.answer(callSid, answerConfig);
           },
           onAnswer);
       } else {
@@ -813,7 +825,10 @@ class Call extends EventEmitter {
     }
 
     this._status = Call.State.Closed;
-    this._mediaHandler.ignore(this.parameters.CallSid);
+    const callSid = this.parameters.CallSid;
+    if (callSid) {
+      this._mediaHandler.ignore(callSid);
+    }
     this._publisher.info('connection', 'ignored-by-local', null, this);
 
     if (this._onIgnore) {
@@ -886,8 +901,11 @@ class Call extends EventEmitter {
     }
 
     this._isRejected = true;
-    this._signalingAdapter.reject(this.parameters.CallSid);
-    this._mediaHandler.reject(this.parameters.CallSid);
+    const callSid = this.parameters.CallSid;
+    if (callSid) {
+      this._signalingAdapter.reject(callSid);
+      this._mediaHandler.reject(callSid);
+    }
     this._publisher.info('connection', 'rejected-by-local', null, this);
     this._cleanupEventListeners();
     this._mediaHandler.close();
@@ -960,8 +978,15 @@ class Call extends EventEmitter {
     this._log.info('Sending digits over PStream');
 
     if (this._signalingAdapter !== null && this._signalingAdapter.status !== 'disconnected') {
+      const callSid = this.parameters.CallSid;
+      if (!callSid) {
+        const error = new GeneralErrors.ConnectionError('Could not send DTMF: CallSid is not set');
+        this._log.debug('#error', error);
+        this.emit('error', error);
+        return;
+      }
       const dtmfConfig: DtmfConfig = { digits };
-      this._signalingAdapter.dtmf(this.parameters.CallSid, dtmfConfig);
+      this._signalingAdapter.dtmf(callSid, dtmfConfig);
     } else {
       const error = new GeneralErrors.ConnectionError('Could not send DTMF: Signaling channel is disconnected');
       this._log.debug('#error', error);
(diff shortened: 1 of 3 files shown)


lib/twilio/call.ts
```diff
@@ -380,10 +380,15 @@ class Call extends EventEmitter {
           this._signalingAdapter.removeListener('hangup', onHangup);
         };
 
+        const callSid = this.parameters.CallSid || this.outboundConnectionId;
+        if (!callSid) {
+          this._log.warn('Cannot reinvite: CallSid and outboundConnectionId are not set');
+          return;
+        }
```
charliesantos, line 387: Isn't this a terminal state so should we throw? When would this happen?

lib/twilio/call.ts
```diff
@@ -665,11 +670,13 @@ class Call extends EventEmitter {
         this._publisher.info('connection', eventName, null, this);
 
         // Report the preferred codec and params as they appear in the SDP
-        const { codecName, codecParams } = getPreferredCodecInfo(this._mediaHandler.version.getSDP());
-        this._publisher.info('settings', 'codec', {
-          codec_params: codecParams,
-          selected_codec: codecName,
-        }, this);
+        if (this._mediaHandler.version) {
```
charliesantos: Same, when would this happen? I would think `version` gets initialized once call object is ready.

lib/twilio/call.ts
```diff
@@ -687,13 +694,22 @@ class Call extends EventEmitter {
       this._signalingAdapter.addListener('hangup', this._onHangup);
 
       if (this._direction === Call.CallDirection.Incoming) {
+        const callSid = this.parameters.CallSid;
+        if (!callSid) {
```
charliesantos: When would this happen?

lib/twilio/call.ts
```diff
@@ -812,8 +828,14 @@ class Call extends EventEmitter {
       return;
     }
 
+    const callSid = this.parameters.CallSid;
+    if (!callSid) {
```
charliesantos: Same. on what instance this will happen?

lib/twilio/call.ts
```diff
@@ -885,9 +907,15 @@ class Call extends EventEmitter {
...
```
charliesantos: Same. on what instance this will happen?

lib/twilio/call.ts
```diff
@@ -960,8 +988,15 @@ class Call extends EventEmitter {
     this._log.info('Sending digits over PStream');
 
     if (this._signalingAdapter !== null && this._signalingAdapter.status !== 'disconnected') {
+      const callSid = this.parameters.CallSid;
+      if (!callSid) {
```
charliesantos: Same. on what instance this will happen?

lib/twilio/call.ts
```diff
@@ -1267,12 +1302,13 @@ class Call extends EventEmitter {
    */
   private _onConnected = (): void => {
     this._log.info('Received connected from pstream');
-    if (this._signalingReconnectToken && this._mediaHandler.version) {
+    const callSid = this.parameters.CallSid;
+    if (this._signalingReconnectToken && this._mediaHandler.version && callSid) {
```
charliesantos, line 1276: Same. on what instance callsid is not available?

lib/twilio/call.ts
```diff
@@ -1368,7 +1404,7 @@ class Call extends EventEmitter {
       return;
     }
 
-    const pc = this._mediaHandler.version.pc;
+    const pc = this._mediaHandler.version?.pc;
```
charliesantos: on what instance `version` is not available but goes to this path?

charliesantos (changes requested):

charliesantos (commented):
I understand the intent. But several code paths leave the call in an inconsistent state. Some paths are effectively dead end/code. Please add comment for each code paths, and make sure the state of the call object is known when these things happen.

lib/twilio/call.ts
kpchoy: removed, should not happen

lib/twilio/call.ts
kpchoy: Added comment

lib/twilio/call.ts
kpchoy: removed, should not happen

lib/twilio/call.ts
kpchoy: removed, should not happen

lib/twilio/call.ts
kpchoy: removed, should not happen

lib/twilio/call.ts
kpchoy: removed, should not happen

lib/twilio/call.ts
kpchoy: added comment

lib/twilio/call.ts
kpchoy: added comment

charliesantos (commented):
should we add tests?

charliesantos (commented):
Comments from Claude. Let me know your thoughts.

### High

1. **`lib/twilio/call.ts:386` — reinvite fallback to `outboundConnectionId` is misleading.**
   `this.parameters.CallSid || this.outboundConnectionId` lets reinvite continue with the temporary local sid. But neither signaling adapter can use that: `SipSignalingAdapter.reinvite()` at `sipsignalingadapter.ts:317-322` calls `_getSession(callSid)` which keys outbound sessions on the *real* CallSid; the pstream path forwards the temp sid to the server where it won't resolve. Either change will silently no-op downstream. The earlier patch version (returning early on `!CallSid`) conveyed the same outcome more clearly.
   Suggestion: revert to `const callSid = this.parameters.CallSid;` and log `warn` on drop, or if the outbound fallback is intentional, document *why* the adapters should accept it (and add a matching adapter path that actually works).

2. **Tests — no coverage added for any of the 4 new guards.** PATCH 5 (the final commit) reverted the tests added in PATCH 1-3. None of the new branches (reinvite early-return, `_onConnected` early-return, `_onMediaFailure` with null `version`, SIP `From` fallback) have regression tests. For a PR whose stated purpose is "add null checks," that's a gap.
   Suggestion: at minimum add one test per guard branch — they're straightforward given existing sinon fixtures.

### Medium

3. **`lib/twilio/call.ts:1381` (`_onMediaFailure`) — optional chaining is a local band-aid.** Making `version?.pc` optional-chained prevents the immediate deref, but the method still falls through to `_mediaReconnectBackoff.backoff()` → `ready` → `iceRestart(…)` which needs `_mediaHandler.version` to exist. If `version` is actually null on entry, the crash just moves a few events later.
   Rationale: `isEndOfIceCycle` is computed from `type` only, so the failure path proceeds without `pc`.
   Suggestion: bail early at the top of `_onMediaFailure` if `!this._mediaHandler.version`, with a warn log.

4. **`lib/twilio/call.ts:386` vs `:1278` — inconsistent treatment of `outboundConnectionId`.** Reinvite accepts the outbound temp sid as a fallback, but `_onConnected`/reconnect does not. Both operations are semantically similar (both require a server-known sid). Pick one policy and document it.
   Suggestion: align both sites on `parameters.CallSid` only, with the outbound-specific case (if real) handled separately.

### Low

5. **`lib/twilio/call.ts:1278` — `_onConnected` guard may be unreachable.** `_signalingReconnectToken` is only set inside `_onAnswer` (`:1228`), which *also* calls `_setCallSid(payload)` (`:1239`) before the caller ever sees the token truthy. So `_signalingReconnectToken && !callSid` is likely impossible in practice. Either clarify the scenario in the comment (and perhaps add a regression test for the actual scenario) or remove the dead check.

6. **`lib/twilio/signaling/sipsignalingadapter.ts:501` — swallowing missing `remoteIdentity`.** Collapsing to empty string hides a likely protocol error — a SIP Invitation without a parseable `remoteIdentity` is malformed. Current consumers don't use `From`, but downstream surface area will.
   Suggestion: log a `warn` when falling back to `''`, so the missing identity appears in debug logs.

7. **PR description drift.** The final PR scope is narrower than the title/body suggest. Update the description to list the 4 final changes so reviewers can cross-check.

### Nit

8. **`call.ts:383-384` comment** — "Both missing means the call is being torn down during ICE restart" overspecifies. Several scenarios hit this path; the comment picks one.

9. **`call.ts:1275-1277` comment** — describes a scenario ("transport reconnects before the initial answer event") that isn't reachable given where `_signalingReconnectToken` is set.

10. **`call.ts:1378-1380` comment** — longer than the code change. Either shorten or link to the reproducer (VBLOCKS-6507).

kpchoy:
### `lib/twilio/call.ts`
  - **Reinvite backoff**: Removed `outboundConnectionId` fallback — now only uses `parameters.CallSid`. The temp outbound sid isn't usable by either adapter downstream.
  - **`_onMediaFailure`**: Changed from optional chaining (`version?.pc`) to an early bail (`if (!version) return`) with warn log, preventing downstream crashes when backoff triggers
  `iceRestart` on a null version.
  - **`_onConnected`**: Kept the `callSid` guard in the existing condition (no behavior change from prior commit).
  - Removed guard comments (bloat)

  ### `lib/twilio/signaling/sipsignalingadapter.ts`
  - Added warn log when incoming INVITE has no parseable `remoteIdentity` URI.

  ### `tests/unit/call.ts`
  - Added test: reinvite skips when CallSid is not set.
  - Added test: `_onConnected` skips reconnect when CallSid is missing.
  - Added test: `_onMediaFailure` does not attempt media reconnect when version is null.

charliesantos (approved):

PR #430, opened 2026-04-30
[VBLOCKS-6372] feat: sdh implementation

kpchoy:
**Contributing to Twilio**

## Pull Request Details

https://twilio-engineering.atlassian.net/browse/VBLOCKS-6372

Successfully tested happy path outgoing call

### Description

  - Replace the no-op SessionDescriptionHandler stub with SipSessionDescriptionHandler, a thin SDP relay that bridges
  SIP.js's Promise-based interface to the SDK's existing PeerConnection via a narrow IPeerConnection contract
  - SipSignalingAdapter registers the real SDH factory and tracks per-call PeerConnection bindings via a
  WeakMap<Session, SdhBinding>; on 200 OK, X-Twilio-CallSid rekeys the outbound session map from temp to server-issued
   sid
  - Call.ts adds a useSip option that skips eager makeOutgoingCall/answerIncomingCall on the SIP path and passes the
  PeerConnection through invite/answer/reconnect configs; onopen decorator fires onAnswer(pc) exactly once when media
  stabilizes
  - Device.SIPSignalingOptions now requires sipUri (SIP identity URI, separate from sipServer transport URL) so the
  SIP registrar can differ from the WebSocket edge host

### Before review
* [ ] Ready for review

### DIFF
--- a/lib/twilio/call.ts
+++ b/lib/twilio/call.ts
@@ -685,24 +685,71 @@ class Call extends EventEmitter {
       }
 
       this._signalingAdapter.addListener('hangup', this._onHangup);
+      this._signalingAdapter.on('answer', this._onAnswer);
 
-      if (this._direction === Call.CallDirection.Incoming) {
+      const callSid = this.parameters.CallSid;
+
+      if (this._options.useSip) {
+        // PeerConnection.onopen fires from multiple stable-state observers,
+        // so we gate onAnswer to a single call.
+        const previousOnOpen = this._mediaHandler.onopen;
+        let onAnswerFired = false;
+        this._mediaHandler.onopen = () => {
+          previousOnOpen();
+          if (!onAnswerFired && this._mediaHandler.version?.pc) {
+            onAnswerFired = true;
+            onAnswer(this._mediaHandler.version.pc);
+          }
+        };
+      }
+
+      const params = Array.from(this.customParameters.entries()).map(pair =>
+        `${encodeURIComponent(pair[0])}=${encodeURIComponent(pair[1])}`).join('&');
+      const outgoingCallSid = this._options.reconnectCallSid || this.outboundConnectionId!;
+
+      const answerViaSip = () => {
         this._isAnswered = true;
-        this._signalingAdapter.on('answer', this._onAnswer);
-        this._mediaHandler.answerIncomingCall(this.parameters.CallSid,
-          this._options.offerSdp, rtcConfiguration,
+        this._signalingAdapter.answer(callSid, {
+          sdp: this._options.offerSdp || '',
+          peerConnection: this._mediaHandler,
+          rtcConfiguration,
+        });
+      };
+
+      const answerViaMediaHandler = () => {
+        this._isAnswered = true;
+        this._mediaHandler.answerIncomingCall(
+          callSid,
+          this._options.offerSdp,
+          rtcConfiguration,
           (answerSdp: string) => {
-            const answerConfig: AnswerConfig = { sdp: answerSdp };
-            this._signalingAdapter.answer(this.parameters.CallSid, answerConfig);
+            this._signalingAdapter.answer(callSid, { sdp: answerSdp });
           },
-          onAnswer);
-      } else {
-        const params = Array.from(this.customParameters.entries()).map(pair =>
-         `${encodeURIComponent(pair[0])}=${encodeURIComponent(pair[1])}`).join('&');
-        this._signalingAdapter.on('answer', this._onAnswer);
+          onAnswer,
+        );
+      };
 
-        const outgoingCallSid = this._options.reconnectCallSid || this.outboundConnectionId!;
+      const inviteViaSip = () => {
+        if (this._signalingReconnectToken) {
+          const reconnectConfig: ReconnectConfig = {
+            sdp: '',
+            reconnectToken: this._signalingReconnectToken,
+            peerConnection: this._mediaHandler,
+            rtcConfiguration,
+          };
+          this._signalingAdapter.reconnect(outgoingCallSid, reconnectConfig);
+        } else {
+          const inviteConfig: InviteConfig = {
+            sdp: '',
+            params,
+            peerConnection: this._mediaHandler,
+            rtcConfiguration,
+          };
+          this._signalingAdapter.invite(outgoingCallSid, inviteConfig);
+        }
+      };
 
+      const inviteViaMediaHandler = () => {
         this._mediaHandler.makeOutgoingCall(outgoingCallSid, rtcConfiguration, (offerSdp: string) => {
           const onPstreamAnswerOrRinging = (payload: any) => {
             if (!payload.sdp) { return; }
@@ -732,7 +779,12 @@ class Call extends EventEmitter {
           // self._setupDTLSTransport in the media handler after this method
           // finishes. See peerconnection.makeOutgoingCall.
         });
+      };
+
+      if (this._direction === Call.CallDirection.Incoming) {
+        return this._options.useSip ? answerViaSip() : answerViaMediaHandler();
       }
+      return this._options.useSip ? inviteViaSip() : inviteViaMediaHandler();
     };
 
     if (this._options.beforeAccept) {
@@ -2139,6 +2191,13 @@ namespace Call {
      * TwiML params for the call. May be set for either outgoing or incoming calls.
      */
     twimlParams?: Record<string, any>;
+
+    /**
+     * When true, the call uses the SIP.js-backed signaling path. The SDP
+     * exchange runs through SipSessionDescriptionHandler instead of being
+     * performed eagerly by Call.
+     */
+    useSip?: boolean;
 
     /**
      * Voice event SID generator.
--- a/lib/twilio/call.ts
+++ b/lib/twilio/call.ts
@@ -189,6 +189,14 @@ class Call extends EventEmitter {
   private _isAnswered: boolean = false;
 
   /**
+   * Whether the SIP-path onopen wrapper has already been installed. Guards
+   * against repeated `accept()` invocations chaining wrappers onto the
+   * previous one (each layer would retain its own onAnswerFired closure
+   * and fire independently against stale state).
+   */
+  private _sipOnOpenWrapped: boolean = false;
+
+  /**
    * Whether the call has been cancelled.
    */
   private _isCancelled: boolean = false;
@@ -689,9 +697,11 @@ class Call extends EventEmitter {
 
       const callSid = this.parameters.CallSid;
 
-      if (this._options.useSip) {
+      if (this._options.useSip && !this._sipOnOpenWrapped) {
         // PeerConnection.onopen fires from multiple stable-state observers,
-        // so we gate onAnswer to a single call.
+        // so we gate onAnswer to a single call. _sipOnOpenWrapped prevents a
+        // repeat accept() from chaining another wrapper over this one.
+        this._sipOnOpenWrapped = true;
         const previousOnOpen = this._mediaHandler.onopen;
         let onAnswerFired = false;
         this._mediaHandler.onopen = () => {
@@ -709,8 +719,9 @@ class Call extends EventEmitter {
 
       const answerViaSip = () => {
         this._isAnswered = true;
+        // SDP intentionally omitted: the SIP adapter derives the answer
+        // from the invitation body via its SDH, not from the caller.
         this._signalingAdapter.answer(callSid, {
-          sdp: this._options.offerSdp || '',
           peerConnection: this._mediaHandler,
           rtcConfiguration,
         });
@@ -739,9 +750,11 @@ class Call extends EventEmitter {
           };
           this._signalingAdapter.reconnect(outgoingCallSid, reconnectConfig);
         } else {
+          // params are intentionally omitted: the SIP adapter does not
+          // currently embed caller-supplied params in the INVITE. If/when
+          // that's added, pass them as extraHeaders through InviteConfig.
           const inviteConfig: InviteConfig = {
             sdp: '',
-            params,
             peerConnection: this._mediaHandler,
             rtcConfiguration,
           };
--- a/lib/twilio/device.ts
+++ b/lib/twilio/device.ts
@@ -1587,9 +1587,16 @@ class Device extends EventEmitter {
       // plain `sip:user@host` or `sip:user@host:port` forms; URIs with
       // parameters (e.g. `;transport=tcp`) would parse incorrectly. Twilio-
       // issued SIP URIs don't carry parameters today — revisit if that changes.
-      const parsedUri = new URL(sipOpts.sipUri.replace(/^sip:/, 'http://'));
+      let parsedUri: URL;
+      try {
+        parsedUri = new URL(sipOpts.sipUri.replace(/^sip:/, 'http://'));
+      } catch {
+        throw new InvalidArgumentError(
+          '`signalingOptions.sipUri` must be a valid SIP URI of the form `sip:user@host` or `sip:user@host:port`',
+        );
+      }
       return new SipSignalingAdapter({
-        sipDomain: parsedUri.hostname,
+        sipDomain: parsedUri.host,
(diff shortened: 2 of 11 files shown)


lib/twilio/signaling/sipsessiondescriptionhandler.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  sendDtmf(_tones: string): boolean {
+    return false;
+  }
+
+  private _awaitOperation<T>(start: (resolve: (value: T) => void) => void): Promise<T> {
+    return new Promise<T>((resolve, reject) => {
+      this._rejectPending = reject;
+      start((value) => {
```
charliesantos, line 144: `_awaitOperation` unconditionally overwrites `this._rejectPending`, so if a second op starts before the first resolves, only the second Promise can be rejected on `onerror` — the first hangs forever. SIP.js's dialog layer can legitimately issue a second `setDescription` (PRACK/UPDATE, early-media, re-negotiation) before the prior Promise settles. Either refuse to start a new op while one is pending, or give every awaiting op its own reject captured via closure and collect them into a `Set` that `_failPending` drains.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -276,6 +276,7 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
     }
     this._pendingInvitations.delete(callSid);
     this._inboundSessions.set(callSid, invitation);
+    this._bindSession(invitation, callSid, config);
     invitation.accept().catch((error: Error) => {
       this._log.error('Failed to accept invitation', error);
       this.emit('error', { error: { code: 31000, message: error.message }, callsid: callSid });
```
charliesantos, line 321: On a failed `invitation.accept()`, `_inboundSessions` and the `_sessionBindings` WeakMap entry still hold the session, and a later `hangup(callSid)` dispatches on `session.state` which may already be `Terminated` (no-op branch) — leaving the map entry to collide with any retry or re-invite for the same CallSid. In this catch handler, also `this._inboundSessions.delete(callSid)` and clear the `_sessionBindings` entry (bindings don't auto-clear since SIP.js internals may still reference the session). Copilot flagged the same thing in their suppressed comment — worth acting on.

lib/twilio/call.ts
```diff
@@ -685,24 +685,71 @@ class Call extends EventEmitter {
...
+      this._signalingAdapter.on('answer', this._onAnswer);
+
+      const callSid = this.parameters.CallSid;
+
+      if (this._options.useSip) {
+        // PeerConnection.onopen fires from multiple stable-state observers,
+        // so we gate onAnswer to a single call.
+        const previousOnOpen = this._mediaHandler.onopen;
+        let onAnswerFired = false;
+        this._mediaHandler.onopen = () => {
```
charliesantos, line 712: This `onopen` wrap will compound if `accept()` runs more than once (error retries, canceled-before-gum re-invocation) — each call chains another `previousOnOpen` and re-captures a fresh `onAnswerFired` flag, leaking closures and potentially firing `onAnswer` against stale state. Consider gating the wrap with an "already wrapped" flag, or installing it inside `connect()` after the `_status === Connecting` re-check so the canceled path never installs it.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -462,6 +468,29 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+      callSid,
+      peerConnection: config.peerConnection,
+      rtcConfiguration: config.rtcConfiguration,
+    });
+  }
+
+  private _getSdhBinding(session: Session): SdhBinding {
+    const binding = this._sessionBindings.get(session);
+    if (!binding) {
+      throw new Error('SipSignalingAdapter: no PeerConnection binding registered for SIP session');
```
charliesantos: Throwing inside SIP.js's `sessionDescriptionHandlerFactory` bubbles through SIP.js's Promise machinery in ways that are hard to reason about — in practice SIP.js treats it as a fatal session error and may not emit a clean `onBye`/`Terminated`. For inbound INVITEs specifically, if SIP.js constructs the SDH eagerly on INVITE receipt (to parse `setDescription(remoteOffer)`), the factory fires before `answer()` registers the binding and this throws. Verify SIP.js's actual call order here; if eager, the inbound path needs a lazy-binding SDH that waits for `answer()` to supply the PeerConnection.

lib/twilio/device.ts
```diff
@@ -1579,13 +1583,24 @@ class Device extends EventEmitter {
...
+        parsedUri = new URL(sipOpts.sipUri.replace(/^sip:/, 'http://'));
+      } catch {
+        throw new InvalidArgumentError(
+          '`signalingOptions.sipUri` must be a valid SIP URI of the form `sip:user@host` or `sip:user@host:port`',
+        );
+      }
       return new SipSignalingAdapter({
-        sipDomain: domain,
-        sipUri: `sip:${sipOpts.sipCredentials.username}@${domain}`,
+        sipDomain: parsedUri.host,
```
charliesantos: Switching to `parsedUri.host` means `sipDomain` now carries `host:port` when a port is present, and it's consumed as `UserAgent.makeURI(\`sip:${sipDomain}\`)` in `_sendInvite` — producing a target like `sip:host:5060`, which SIP.js may parse as a hostport but is unconventional (Twilio's stack typically wants `sip:host` with transport in `transportOptions`). If Twilio-issued SIP URIs never carry ports (per the comment above), then `host === hostname` and the change wasn't needed; if they can, verify the port-in-domain form doesn't produce malformed Route or Request-URI headers.

lib/twilio/signaling/sipsessiondescriptionhandler.ts
```diff
@@ -0,0 +1,144 @@
...
+        this._rtcConfiguration,
+        (answerSdp) => { this._cachedAnswer = answerSdp; },
+        () => resolve(),
+      );
+    });
+  }
+
+  close(): void {
+    this._failPending(new Error('SipSessionDescriptionHandler closed'));
+    this._pc.close();
```
charliesantos: SIP.js calls `sdh.close()` during session teardown, which here calls `this._pc.close()` — but Call also closes the PeerConnection via `_cleanupEventListeners`/`_disconnect`, so `close()` may run twice. Even if `PeerConnection.close` is idempotent, double-close risks duplicate `onclose` events or tripping noop-log warnings on torn-down handlers. Consider making SDH.close a no-op for the PC (Call owns its lifecycle, SDH is borrowed) and documenting that inversion.

lib/twilio/signaling/sipsessiondescriptionhandler.ts
```diff
@@ -0,0 +1,144 @@
...
+  // here; the onerror hook in the constructor triggers it.
+  private _rejectPending: ((error: Error) => void) | null = null;
+
+  constructor(
+    private _pc: IPeerConnection,
+    private _callSid: string,
+    private _rtcConfiguration: RTCConfiguration = {},
+  ) {
+    const previousOnError = this._pc.onerror;
+    this._pc.onerror = (error: PeerConnectionError) => {
```
charliesantos, line 79: This wrap relies on Call having already assigned `_mediaHandler.onerror` by the time the SDH is constructed — currently true because the factory runs inside SIP.js's `invite()`/`answer()`, which is after Call's `accept()` wires its handler. Brittle: a future refactor that constructs the SDH earlier would silently break error propagation. Add an integration-style test that asserts `Call.onerror` fires end-to-end when the SDH is created after Call's `onerror` is assigned (the existing sinon-spy test doesn't cover this ordering contract across Call+Adapter+SDH).

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -540,8 +574,17 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
         onProgress: () => {
           this.emit('ringing', { callsid: callSid });
         },
-        onAccept: () => {
-          const payload: Record<string, any> = { callsid: callSid, sdp: '' };
+        onAccept: (response: any) => {
+          const serverCallSid = response?.message?.getHeader?.('X-Twilio-CallSid');
+          if (serverCallSid && serverCallSid !== callSid) {
+            this._outboundSessions.delete(callSid);
+            this._outboundSessions.set(serverCallSid, inviter);
```
charliesantos, line 616: The rekey updates `_outboundSessions` and the bindings WeakMap lookup still works (keyed by Session), but the binding's `callSid` field still holds the old temp sid. Any future reader of `binding.callSid` (debug logging, SDH refactors) would see a stale identifier. Either update `binding.callSid = serverCallSid` in this block, or document that `binding.callSid` is intentionally the initially-bound temp sid.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -334,8 +335,13 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
     });
   }
 
-  reconnect(callSid: string, { sdp, reconnectToken }: ReconnectConfig): void {
-    this._sendInvite(sdp, callSid, undefined, reconnectToken);
+  reconnect(callSid: string, config: ReconnectConfig): void {
+    this._sendInvite(callSid, {
+      sdp: config.sdp,
+      params: '',
```
charliesantos: `reconnect` hardcodes `params: ''` when forwarding to `_sendInvite`, silently substituting for the required `params: string` on `InviteConfig`. Consider making `params` optional on `InviteConfig` (defaulting to `''`) to reflect reality, or adding a comment noting that reconnect intentionally drops params (the pstream reconnect path carries none).

lib/twilio/call.ts
```diff
@@ -685,24 +685,71 @@ class Call extends EventEmitter {
...
+        `${encodeURIComponent(pair[0])}=${encodeURIComponent(pair[1])}`).join('&');
+      const outgoingCallSid = this._options.reconnectCallSid || this.outboundConnectionId!;
+
+      const answerViaSip = () => {
         this._isAnswered = true;
-        this._signalingAdapter.on('answer', this._onAnswer);
-        this._mediaHandler.answerIncomingCall(this.parameters.CallSid,
-          this._options.offerSdp, rtcConfiguration,
+        this._signalingAdapter.answer(callSid, {
+          sdp: this._options.offerSdp || '',
```
charliesantos: `offerSdp || ''` coerces a missing SDP into dead data — the SIP adapter ignores `config.sdp` and re-derives SDP from the invitation body via the SDH, but `AnswerConfig.sdp` is a required `string`. Either make `sdp` optional on `AnswerConfig` for the SIP branch, or assert non-empty here; the silent empty-string coerces a real misconfiguration into a hard-to-trace bug.

lib/twilio/signaling/sipsessiondescriptionhandler.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  getDescription(): Promise<BodyAndContentType> {
+    if (this._cachedAnswer !== null) {
+      const body = this._cachedAnswer;
+      this._cachedAnswer = null;
+      return Promise.resolve({ body, contentType: APPLICATION_SDP });
+    }
+    return this._awaitOperation<BodyAndContentType>((resolve) => {
+      this._pc.makeOutgoingCall(this._callSid, this._rtcConfiguration, (offerSdp) => {
+        this._hasSentOffer = true;
```
charliesantos, line 93: `_hasSentOffer` is set to `true` but never reset, so this SDH can never serve an inbound answer afterward. For re-INVITEs within a session, SIP.js often constructs a fresh SDH — but if it ever reuses this one, subsequent `setDescription` will wrongly go down the `processAnswer` branch even on an inbound re-INVITE. Either reset after a full offer/answer exchange completes, or document "SDH is single-use per offer/answer exchange" and verify SIP.js constructs a new one per negotiation.

lib/twilio/signaling/sipsessiondescriptionhandler.ts
```diff
@@ -0,0 +1,144 @@
...
+    });
+  }
+
+  close(): void {
+    this._failPending(new Error('SipSessionDescriptionHandler closed'));
+    this._pc.close();
+  }
+
+  sendDtmf(_tones: string): boolean {
+    return false;
```
charliesantos, line 138: Returning `false` tells SIP.js "I don't implement DTMF" — correct, since Twilio's DTMF flow goes through `adapter.dtmf()` (SIP INFO). A one-line comment saying exactly that would save the next reader a trace through the signaling adapter.

lib/twilio/device.ts
```diff
@@ -1579,13 +1583,24 @@ class Device extends EventEmitter {
...
       this._log.info('Setting up SIP signaling');
       const sipOpts = this._options.signalingOptions as Device.SIPSignalingOptions;
-      const domain = new URL(sipOpts.sipServer).hostname;
+      // Reuse the browser's URL parser by swapping the scheme. Only safe for
+      // plain `sip:user@host` or `sip:user@host:port` forms; URIs with
+      // parameters (e.g. `;transport=tcp`) would parse incorrectly. Twilio-
+      // issued SIP URIs don't carry parameters today — revisit if that changes.
+      let parsedUri: URL;
+      try {
+        parsedUri = new URL(sipOpts.sipUri.replace(/^sip:/, 'http://'));
```
charliesantos: `SIP:alice@host` (uppercase scheme, valid per RFC 3261) won't match `/^sip:/` and `new URL` will throw into the catch — a clear error surfaces now, so this is self-documenting. Still worth adopting Copilot's case-insensitive + `sips:` variant if `sips:` URIs are plausible; otherwise leave as-is.

lib/twilio/call.ts
```diff
@@ -685,24 +685,71 @@ class Call extends EventEmitter {
...
+            sdp: '',
+            reconnectToken: this._signalingReconnectToken,
+            peerConnection: this._mediaHandler,
+            rtcConfiguration,
+          };
+          this._signalingAdapter.reconnect(outgoingCallSid, reconnectConfig);
+        } else {
+          const inviteConfig: InviteConfig = {
+            sdp: '',
+            params,
```
charliesantos: `params` is computed unconditionally but the SIP adapter doesn't embed custom params in the INVITE, so this value is effectively dead on the SIP branch. Either wire it into the SIP INVITE (e.g., as headers — the non-SIP path uses a URL-encoded body), or pass `params: ''` on this branch and drop the computation.

tests/unit/sipsignalingadapter.ts
```diff
@@ -2,6 +2,19 @@ import * as assert from 'assert';
...
+  return {
+    onerror: () => { /* replaced by SDH constructor when used */ },
+    makeOutgoingCall: sinon.stub(),
+    answerIncomingCall: sinon.stub(),
+    processAnswer: sinon.stub(),
+    close: sinon.stub(),
+  };
+}
+
+const pc = (): IPeerConnection => createPeerConnectionStub();
```
charliesantos: This alias only saves keystrokes. Inlining `createPeerConnectionStub()` at call sites is more searchable.

charliesantos (changes requested):
Solid structural work — the SDH is a thin, well-scoped bridge and the rekey-on-X-Twilio-CallSid logic is clean. A few real bugs in the error-handling glue stand out (single `_rejectPending` slot, stale `_inboundSessions` after a failed `invitation.accept()`, throwing inside SIP.js's SDH factory). Worth addressing before the feature branch lands on `master`.

On error code 31000 on SIP accept failure (`sipsignalingadapter.ts`): 31000 is "general" — consider wrapping the caught error through `getPreciseSignalingErrorByCode` (already used elsewhere in `call.ts`) when the underlying error carries a SIP status code.

On test coverage: no test exercises two SDH operations in flight simultaneously (H1), and no test covers the inbound flow where the SDH factory runs before `answer()` supplies the binding (H4). The `SipSessionDescriptionHandler` suite is otherwise thorough.

(1 finding from the local review was omitted as already raised by @copilot-pull-request-reviewer.)

charliesantos (approved):

PR #432, opened 2026-05-06
[VBLOCKS-6375] feat: sdh ice restart support

kpchoy:
**Contributing to Twilio**

## Pull Request Details

https://twilio-engineering.atlassian.net/browse/VBLOCKS-6375

### Description

  - **Add SDH ICE restart support** so the SIP signaling path can recover media via re-INVITE with fresh ICE, matching the existing PStream behavior
  - **Replace `reinvite()` with `iceRestart()`** on `SignalingAdapter` — each adapter now owns its offer-generation strategy:
    - `PStreamSignalingAdapter`: asks `mediaHandler.iceRestart()` for the offer, ships it via `pstream.reinvite`
    - `SipSignalingAdapter`: calls `session.invite()` with `offerOptions.iceRestart`, letting SIP.js drive the SDH to produce the offer
  - **Wire ICE restart through `SipSessionDescriptionHandler`** — `getDescription({offerOptions:{iceRestart:true}})` routes to `pc.iceRestart()` instead of `pc.makeOutgoingCall()`, and the following `setDescription` goes through
  `processAnswer` (outbound-offer semantics)
  - **Wrap `pc.onfailed`** in the SDH, scoped by an `_iceRestartPending` flag, so `createOffer` rejections reject the pending Promise without breaking unrelated in-flight `setDescription` Promises on runtime ICE failures
  - **Simplify `Call._mediaReconnectBackoff` handler** — delegates offer generation to the adapter and relies on adapter-emitted `hangup` for cleanup on failure
  - **SIP failure paths all emit `hangup`** (no session, `onReject`, async `.catch`) so `Call` can tear down its re-INVITE listeners and exit the reconnecting state

### Before review
* [ ] Ready for review

### DIFF
--- a/lib/twilio/call.ts
+++ b/lib/twilio/call.ts
@@ -376,6 +376,10 @@ class Call extends EventEmitter {
       }
 
       const onAnswerOrRinging = (payload: any) => {
+        // The adapter is shared across concurrent calls; ignore events for
+        // other calls so their answer/hangup doesn't abort this call's
+        // in-flight ICE restart by removing the listeners below.
+        if (payload?.callsid !== callSid) { return; }
         this._signalingAdapter.removeListener('answer', onAnswerOrRinging);
         this._signalingAdapter.removeListener('hangup', onHangup);
 
@@ -391,7 +395,8 @@ class Call extends EventEmitter {
         this._mediaHandler.processAnswer(payload.sdp, () => { /* noop for ICE restart */ });
       };
 
-      const onHangup = () => {
+      const onHangup = (payload: any) => {
+        if (payload?.callsid !== callSid) { return; }
         this._signalingAdapter.removeListener('answer', onAnswerOrRinging);
         this._signalingAdapter.removeListener('hangup', onHangup);
       };
--- a/lib/twilio/call.ts
+++ b/lib/twilio/call.ts
@@ -8,7 +8,6 @@ import {
   HangupConfig,
   InviteConfig,
   ReconnectConfig,
-  ReinviteConfig,
   SendMessageConfig,
   SignalingAdapter,
 } from './signaling/signalingadapter';
@@ -370,34 +369,39 @@ class Call extends EventEmitter {
 
     this._mediaReconnectBackoff = new Backoff(BACKOFF_CONFIG);
     this._mediaReconnectBackoff.on('ready', () => {
-      this._mediaHandler.iceRestart((offerSdp: string) => {
-        const onAnswerOrRinging = (payload: any) => {
-          this._signalingAdapter.removeListener('answer', onAnswerOrRinging);
-          this._signalingAdapter.removeListener('hangup', onHangup);
+      const callSid = this.parameters.CallSid;
+      if (!callSid) {
+        this._log.warn('Cannot reinvite: CallSid is not set');
+        return;
+      }
 
-          if (!payload.sdp || !this._mediaHandler.version
-              || this._mediaHandler.version.pc.signalingState !== 'have-local-offer') {
-            return;
-          }
+      const onAnswerOrRinging = (payload: any) => {
+        this._signalingAdapter.removeListener('answer', onAnswerOrRinging);
+        this._signalingAdapter.removeListener('hangup', onHangup);
 
-          this._mediaHandler.processAnswer(payload.sdp, () => { /* noop for ICE restart */ });
-        };
-
-        const onHangup = () => {
-          this._signalingAdapter.removeListener('answer', onAnswerOrRinging);
-          this._signalingAdapter.removeListener('hangup', onHangup);
-        };
-
-        const callSid = this.parameters.CallSid;
-        if (!callSid) {
-          this._log.warn('Cannot reinvite: CallSid is not set');
+        // On the SIP path, the remote answer reaches PeerConnection via
+        // the SDH before the adapter emits 'answer', so payload.sdp is
+        // empty — the guard below skips a redundant processAnswer. On
+        // the PStream path, payload.sdp carries the real answer SDP.
+        if (!payload.sdp || !this._mediaHandler.version
+            || this._mediaHandler.version.pc.signalingState !== 'have-local-offer') {
           return;
         }
-        this._signalingAdapter.on('answer', onAnswerOrRinging);
-        this._signalingAdapter.on('hangup', onHangup);
-        const reinviteConfig: ReinviteConfig = { sdp: offerSdp };
-        this._signalingAdapter.reinvite(callSid, reinviteConfig);
-      });
+
+        this._mediaHandler.processAnswer(payload.sdp, () => { /* noop for ICE restart */ });
+      };
+
+      const onHangup = () => {
+        this._signalingAdapter.removeListener('answer', onAnswerOrRinging);
+        this._signalingAdapter.removeListener('hangup', onHangup);
+      };
+
+      this._signalingAdapter.on('answer', onAnswerOrRinging);
+      this._signalingAdapter.on('hangup', onHangup);
+      // Adapter encapsulates its own offer-generation strategy. On
+      // failure the adapter emits 'hangup' for this callSid so the
+      // cleanup listeners above run.
+      this._signalingAdapter.iceRestart(callSid, { mediaHandler: this._mediaHandler });
     });
 
     // temporary call sid to be used for outgoing calls
--- a/lib/twilio/signaling/pstreamsignalingadapter.ts
+++ b/lib/twilio/signaling/pstreamsignalingadapter.ts
@@ -83,9 +83,13 @@ export class PStreamSignalingAdapter extends EventEmitter implements SignalingAd
   iceRestart(callSid: string, { mediaHandler }: IceRestartConfig): void {
     // Call (on PStream) owns offer generation: mediaHandler produces a
     // fresh-ICE offer; we ship it via the existing reinvite wire.
-    mediaHandler.iceRestart((offerSdp: string) => {
-      this._pstream.reinvite(offerSdp, callSid);
-    });
+    try {
+      mediaHandler.iceRestart((offerSdp: string) => {
+        this._pstream.reinvite(offerSdp, callSid);
+      });
+    } catch {
+      this.emit('hangup', { callsid: callSid });
+    }
   }
 
   reconnect(callSid: string, { sdp, reconnectToken }: ReconnectConfig): void {
--- a/lib/twilio/signaling/pstreamsignalingadapter.ts
+++ b/lib/twilio/signaling/pstreamsignalingadapter.ts
@@ -3,9 +3,9 @@ import {
   AnswerConfig,
   DtmfConfig,
   HangupConfig,
+  IceRestartConfig,
   InviteConfig,
   ReconnectConfig,
-  ReinviteConfig,
   SendMessageConfig,
   SignalingAdapter,
   SignalingAdapterStatus,
@@ -80,8 +80,12 @@ export class PStreamSignalingAdapter extends EventEmitter implements SignalingAd
     this._pstream.reject(callSid);
   }
 
-  reinvite(callSid: string, { sdp }: ReinviteConfig): void {
-    this._pstream.reinvite(sdp, callSid);
+  iceRestart(callSid: string, { mediaHandler }: IceRestartConfig): void {
+    // Call (on PStream) owns offer generation: mediaHandler produces a
+    // fresh-ICE offer; we ship it via the existing reinvite wire.
+    mediaHandler.iceRestart((offerSdp: string) => {
+      this._pstream.reinvite(offerSdp, callSid);
+    });
   }
 
   reconnect(callSid: string, { sdp, reconnectToken }: ReconnectConfig): void {
--- a/lib/twilio/signaling/signalingadapter.ts
+++ b/lib/twilio/signaling/signalingadapter.ts
@@ -26,8 +26,19 @@ export interface HangupConfig {
   message?: string;
 }
 
-export interface ReinviteConfig {
-  sdp: string;
+/**
+ * Narrow view of the media handler that SignalingAdapter.iceRestart()
+ * needs. Satisfied structurally by PeerConnection.
+ */
+export interface IMediaHandler {
+  iceRestart(onOfferReady: (offerSdp: string) => void): void;
+}
+
+export interface IceRestartConfig {
+  // Used by the PStream adapter to generate the fresh-ICE offer; the
+  // SIP adapter ignores it because SIP.js asks the SDH to produce the
+  // offer via session.invite() + offerOptions.iceRestart.
+  mediaHandler: IMediaHandler;
 }
 
 export interface ReconnectConfig {
@@ -63,8 +74,19 @@ export interface SignalingAdapter extends EventEmitter {
   answer(callSid: string, config: AnswerConfig): void;
   hangup(callSid: string, config: HangupConfig): void;
   reject(callSid: string): void;
-  reinvite(callSid: string, config: ReinviteConfig): void;
   reconnect(callSid: string, config: ReconnectConfig): void;
+
+  /**
+   * Trigger an ICE restart for the given call. Each adapter encapsulates
+   * its own offer-generation strategy:
+   *  - PStream: asks config.mediaHandler to generate the offer, then
+   *    ships it via the existing reinvite wire.
+   *  - SIP: calls session.invite() with offerOptions.iceRestart so SIP.js
+   *    asks the SDH to produce the offer itself.
+   * On failure, emits 'hangup' for the callSid so Call can clean up its
+   * reinvite listeners.
+   */
+  iceRestart(callSid: string, config: IceRestartConfig): void;
 
   dtmf(callSid: string, config: DtmfConfig): void;
   sendMessage(callSid: string, config: SendMessageConfig): void;
--- a/lib/twilio/signaling/signalingadapter.ts
+++ b/lib/twilio/signaling/signalingadapter.ts
@@ -26,19 +26,12 @@ export interface HangupConfig {
   message?: string;
 }
 
-/**
- * Narrow view of the media handler that SignalingAdapter.iceRestart()
- * needs. Satisfied structurally by PeerConnection.
- */
-export interface IMediaHandler {
-  iceRestart(onOfferReady: (offerSdp: string) => void): void;
-}
-
 export interface IceRestartConfig {
   // Used by the PStream adapter to generate the fresh-ICE offer; the
   // SIP adapter ignores it because SIP.js asks the SDH to produce the
-  // offer via session.invite() + offerOptions.iceRestart.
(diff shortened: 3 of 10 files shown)


lib/twilio/call.ts
```diff
@@ -370,34 +369,39 @@ class Call extends EventEmitter {
...
-        this._signalingAdapter.on('answer', onAnswerOrRinging);
-        this._signalingAdapter.on('hangup', onHangup);
-        const reinviteConfig: ReinviteConfig = { sdp: offerSdp };
-        this._signalingAdapter.reinvite(callSid, reinviteConfig);
-      });
+
+        this._mediaHandler.processAnswer(payload.sdp, () => { /* noop for ICE restart */ });
+      };
+
+      const onHangup = () => {
```
charliesantos: `onHangup` doesn't filter on `payload.callsid`, so a concurrent call's hangup on this adapter would silently abort this ICE restart — now more likely since the SIP adapter emits `hangup` in three new paths. Gate both handlers with `if (payload?.callsid !== callSid) return;` before removing listeners.

lib/twilio/signaling/sipsessiondescriptionhandler.ts
```diff
@@ -53,23 +75,42 @@ const APPLICATION_SDP = 'application/sdp';
...
  */
 export class SipSessionDescriptionHandler implements SessionDescriptionHandler {
   private _cachedAnswer: string | null = null;
   private _hasSentOffer: boolean = false;
+  // True while an ICE-restart getDescription() is in flight. Scopes the
+  // onfailed wrap: pc.onfailed fires both for createOffer rejection (what
+  // we want to reject the pending Promise for) and for ICE state = failed
+  // at runtime (what we do NOT want to reject unrelated in-flight
+  // setDescription Promises for).
+  private _iceRestartPending: boolean = false;
```
charliesantos, line 102: `_iceRestartPending` is a single boolean, so two overlapping ICE-restart `getDescription` calls would let the first completion clear the flag and silently swallow a subsequent `onfailed`. Hard to trigger today given serial backoff, but the invariant isn't defended. Either document "at most one concurrent ICE-restart getDescription" or key a Set off the pending resolver.

lib/twilio/signaling/signalingadapter.ts
```diff
@@ -26,8 +26,19 @@ export interface HangupConfig {
   message?: string;
 }
 
-export interface ReinviteConfig {
-  sdp: string;
+/**
+ * Narrow view of the media handler that SignalingAdapter.iceRestart()
+ * needs. Satisfied structurally by PeerConnection.
+ */
+export interface IMediaHandler {
```
charliesantos: `IMediaHandler` is only used by the PStream adapter (SIP adapter ignores `_config`), so placing it in the shared interface couples the base to one implementation's detail. Move it to `pstreamsignalingadapter.ts` and import from `IceRestartConfig`, or extract to a small `mediahandler.ts`.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -354,23 +354,49 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
-      this._log.warn('reinvite: no established session for callSid', callSid);
+      this._log.warn('iceRestart: no established session for callSid', callSid);
+      this.emit('hangup', {
+        callsid: callSid,
+        error: { code: 31000, message: 'No established session for ICE restart' },
+      });
       return;
     }
-    session.invite({
+    const inviteOptions: any = {
```
charliesantos: `inviteOptions: any` bypasses typechecking on `offerOptions.iceRestart` — typos won't fail the build. Import the SDH's extended `SessionDescriptionHandlerOptions`, or inline a minimal structural type.

lib/twilio/signaling/sipsessiondescriptionhandler.ts
```diff
@@ -80,14 +121,45 @@ export class SipSessionDescriptionHandler implements SessionDescriptionHandler {
       this._failPending(error?.info?.twilioError || new Error(error?.info?.message || 'PeerConnection error'));
```
charliesantos, line 129: When `onerror` fires mid-ICE-restart, `_failPending` rejects but `_iceRestartPending` stays `true`, so a later runtime `onfailed` would re-enter `_failPending` — a no-op today but semantically wrong. Clear `_iceRestartPending = false` in the `onerror` wrapper right after `_failPending`.

tests/unit/call.ts
```diff
@@ -2772,12 +2772,33 @@ describe('Call', function() {
...
+      assert.strictEqual(pstream.iceRestart.firstCall.args[0], 'CAtest');
+      assert.deepStrictEqual(pstream.iceRestart.firstCall.args[1], { mediaHandler: conn['_mediaHandler'] });
+    });
+
+    it('self-cleans answer/hangup listeners when adapter emits hangup for the callSid (re-INVITE failure path)', () => {
+      conn = new Call(config, Object.assign(options));
+      conn.parameters = { CallSid: 'CAtest' };
+      pstream.iceRestart = sinon.stub();
+      conn['_mediaReconnectBackoff'].emit('ready');
+      const before = pstream.listenerCount('answer') + pstream.listenerCount('hangup');
```
charliesantos: Aggregate `listenerCount('answer') + listenerCount('hangup')` could mask a leak where one listener is added as another is removed. Worth considering asserting each event's count is 0 independently after the hangup.

lib/twilio/signaling/sipsessiondescriptionhandler.ts
```diff
@@ -53,23 +75,42 @@ const APPLICATION_SDP = 'application/sdp';
...
+ * reject unrelated in-flight setDescription Promises.
  */
 export class SipSessionDescriptionHandler implements SessionDescriptionHandler {
   private _cachedAnswer: string | null = null;
   private _hasSentOffer: boolean = false;
+  // True while an ICE-restart getDescription() is in flight. Scopes the
+  // onfailed wrap: pc.onfailed fires both for createOffer rejection (what
+  // we want to reject the pending Promise for) and for ICE state = failed
+  // at runtime (what we do NOT want to reject unrelated in-flight
+  // setDescription Promises for).
```
charliesantos: Nit: docblock says "setDescription Promises" but the in-flight Promise at this point is a `getDescription` (ICE-restart offer). Worth tweaking.

charliesantos (changes requested):
Solid design — adapter-owned offer generation is the right abstraction. A few items worth discussing inline before merge.

lib/twilio/signaling/signalingadapter.ts
kpchoy: Decided to inline the structural type. Deletes cleanly when PStream goes away. No need for another file.

charliesantos: PStream will not go away for a very long time.

kpchoy: In that case implemented the second option.  Extracted to a smaller `mediahanlder.ts`

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -354,23 +364,49 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+   * listeners and transition out of the reconnecting state.
+   */
+  iceRestart(callSid: string, _config: IceRestartConfig): void {
     const session = this._getSession(callSid);
     if (!session || session.state !== SessionState.Established) {
-      this._log.warn('reinvite: no established session for callSid', callSid);
+      this._log.warn('iceRestart: no established session for callSid', callSid);
+      this.emit('hangup', {
+        callsid: callSid,
+        error: { code: 31000, message: 'No established session for ICE restart' },
```
charliesantos: Emitting `hangup` with an `error` field on the no-session path will route through `Call._onHangup` → `emit('error', GeneralErrors.ConnectionError)` whenever the backoff fires before Call has transitioned to Closed — the old `reinvite()` path stayed silent here. Drop the `error` field from this branch (so Call cleans up its re-INVITE listeners and exits Reconnecting without surfacing a ConnectionError), or raise a distinct signal. Is the customer-visible error intentional?

lib/twilio/signaling/sipsessiondescriptionhandler.ts
```diff
@@ -53,41 +75,99 @@ const APPLICATION_SDP = 'application/sdp';
...
+      // Only reject pending Promises if an ICE restart is in flight —
+      // that's the one case where pc.onfailed carries a createOffer
+      // rejection we need to propagate. Runtime ICE failures (ICE state
+      // = failed with no restart requested) must not reject unrelated
+      // setDescription Promises.
+      if (this._iceRestartPending) {
+        this._iceRestartPending = false;
+        this._failPending(new Error(message || 'PeerConnection failed'));
+      }
+      previousOnFailed(message);
```
charliesantos, line 149: When `_iceRestartPending` is true, `_failPending` already translates the createOffer rejection into a hangup via SIP.js's rejected `session.invite()` → adapter `.catch` → `emit('hangup')`. Calling `previousOnFailed(message)` after that double-dispatches: Call's `_onMediaFailure(ConnectionFailed)` runs concurrently with `_onHangup`. Skip `previousOnFailed` when `_iceRestartPending` was true, so the event is delivered once through the adapter path.

tests/unit/sipsessiondescriptionhandler.ts
```diff
@@ -75,6 +80,115 @@ describe('SipSessionDescriptionHandler', () => {
...
+        () => { settled = true; },
+      );
+      // Simulate runtime ICE failure (no ICE restart requested).
+      pc.onfailed('ICE connection failed');
+      // Yield a microtask or two to let any spurious reject propagate.
+      await Promise.resolve();
+      await Promise.resolve();
+      assert.strictEqual(settled, false);
+      // Clean up the hung Promise so mocha doesn't warn about it.
+      void pending;
```
charliesantos: `void pending;` leaves the Promise permanently in `_pendingRejects` — if a future bug leaked rejects across tests, this test wouldn't surface it. Worth ending with `handler.close(); await Promise.resolve(); assert.strictEqual(settled, true);` — cleans up the hanging Promise and exercises the close() reject path.

charliesantos (changes requested):
Prior review asks are all addressed; flagging two novel behavioral notes on SIP failure-path dispatch and one test-hygiene nit — none blocking.

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: removed the error field.

lib/twilio/signaling/pstreamsignalingadapter.ts
```diff
@@ -80,8 +80,12 @@ export class PStreamSignalingAdapter extends EventEmitter implements SignalingAd
     this._pstream.reject(callSid);
   }
 
-  reinvite(callSid: string, { sdp }: ReinviteConfig): void {
-    this._pstream.reinvite(sdp, callSid);
+  iceRestart(callSid: string, { mediaHandler }: IceRestartConfig): void {
+    // Call (on PStream) owns offer generation: mediaHandler produces a
+    // fresh-ICE offer; we ship it via the existing reinvite wire.
+    mediaHandler.iceRestart((offerSdp: string) => {
```
charliesantos: [M1] `mediaHandler.iceRestart(cb)` has no try/catch and no fallback emit: if it throws synchronously or the callback never fires, Call's inline `answer`/`hangup` listeners accumulate across backoff attempts for the duration of the call. Wrap in try/catch and emit `this.emit('hangup', { callsid: callSid })` on failure, mirroring the SIP adapter's no-session path.

tests/unit/call.ts
```diff
@@ -2772,12 +2772,40 @@ describe('Call', function() {
...
+      // Capture per-event baselines BEFORE the backoff adds its listeners so we
+      // can assert each event's count returns to its baseline after hangup (an
+      // aggregate sum could mask a leak where a listener is added as another
+      // is removed).
+      const answerBaseline = pstream.listenerCount('answer');
+      const hangupBaseline = pstream.listenerCount('hangup');
+      conn['_mediaReconnectBackoff'].emit('ready');
+      assert.strictEqual(pstream.listenerCount('answer'), answerBaseline + 1);
+      assert.strictEqual(pstream.listenerCount('hangup'), hangupBaseline + 1);
+      pstream.emit('hangup', { callsid: 'CAtest', error: { code: 488, message: 'rejected' } });
```
charliesantos: [M2] Emitting `hangup` with an `error` field routes through `Call._onHangup` → `emit('error', ConnectionError)`. If the test Call has no `'error'` listener, Node's EventEmitter throws. Either register `conn.on('error', () => {})` before this emit, or drop the `error` field — the test's intent is listener-count cleanup, not error propagation.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -354,23 +364,46 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
         onAccept: () => {
           this.emit('answer', { callsid: callSid, sdp: '' });
         },
         onReject: (response: any) => {
-          this._log.warn('re-INVITE rejected', response?.message?.statusCode);
+          const code = response?.message?.statusCode;
+          this._log.warn('ICE-restart re-INVITE rejected', code);
+          this.emit('hangup', {
+            callsid: callSid,
+            error: { code: code || 31000, message: `ICE-restart re-INVITE rejected (${code || 'unknown'})` },
```
charliesantos: [L1] `code || 31000` replaces a statusCode of `0` (non-standard but possible) with 31000. Prefer `code ?? 31000` — nullish coalescing falls back only on `null`/`undefined`. Applies to both the `error.code` assignment and the message template.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -354,23 +364,46 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
-      this._log.warn('Failed to send re-INVITE', error);
+      sessionDescriptionHandlerOptions: {
+        offerOptions: { iceRestart: true },
+      },
+    };
+    session.invite(inviteOptions).catch((error: Error) => {
+      this._log.warn('Failed to send ICE-restart re-INVITE', error);
+      this.emit('hangup', {
+        callsid: callSid,
+        error: { code: 31000, message: `Failed to send ICE-restart re-INVITE: ${error.message}` },
```
charliesantos, line 409: [L2] `message: 'Failed to send ICE-restart re-INVITE: ' + error.message` interpolates untrusted SIP.js content (potentially including remote SDP fragments) into the emitted payload. No DOM blast radius here, but consumers who log or surface this see server-controlled strings. Worth truncating `error.message` or documenting that callers must treat it as untrusted.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -354,23 +364,46 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
-      this._log.warn('reinvite: no established session for callSid', callSid);
+      this._log.warn('iceRestart: no established session for callSid', callSid);
+      this.emit('hangup', { callsid: callSid });
       return;
     }
-    session.invite({
+    const inviteOptions: IceRestartInviteOptions = {
       requestDelegate: {
         onAccept: () => {
           this.emit('answer', { callsid: callSid, sdp: '' });
```
charliesantos, line 390: [L3] `onAccept` emits `sdp: ''` by design — `call.ts`'s `onAnswerOrRinging` skips `processAnswer` on empty SDP because the SDH consumed the remote answer via SIP.js first. Worth adding an inline comment here (`// SDP intentionally empty: the SDH consumed the remote answer before this event fires.`); a future reader plugging in a real SDP would silently trigger double-processAnswer.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -354,23 +364,46 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+  /**
+   * Trigger an ICE restart by sending a re-INVITE with
+   * offerOptions.iceRestart:true. SIP.js asks the SDH to produce the
+   * fresh-ICE offer via getDescription(options), so config.mediaHandler
+   * is unused here (the PStream adapter needs it; we preserve the shared
+   * interface). All failure paths (no session, onReject, catch) emit
+   * 'hangup' for the callSid so Call can clean up its inline re-INVITE
+   * listeners and transition out of the reconnecting state.
+   */
+  iceRestart(callSid: string, _config: IceRestartConfig): void {
```
charliesantos, line 376: [N1] `_config` is unused on the SIP path — the `_` prefix is the right TS idiom and the docblock explains why. Worth considering a discriminated union for `IceRestartConfig` long-term so each adapter declares only the fields it needs.

tests/unit/sipsessiondescriptionhandler.ts
```diff
@@ -75,6 +80,129 @@ describe('SipSessionDescriptionHandler', () => {
...
+
+    it('rejects the pending getDescription if pc.onfailed fires during ICE restart (createOffer rejection path)', async () => {
+      const pc = createPeerConnectionStub(); // iceRestart stub never calls back
+      const { handler } = createHandler(pc);
+      const pending = (handler.getDescription as any)(ICE_RESTART_OPTS);
+      pc.onfailed('createOffer rejected');
+      await assert.rejects(pending, /createOffer rejected/);
+    });
+
+    it('still invokes the previous onfailed handler when pc.onfailed fires', () => {
```
charliesantos, line 165: [N2] This test name covers only the no-ICE-restart branch of the `onfailed` wrap. Worth adding a brief inline comment noting `_iceRestartPending` is false here — otherwise a reader might mistake this for the double-dispatch case.

charliesantos (changes requested):
Sorry — I missed a couple of items in the earlier rounds; flagging them now. Two mediums worth fixing before merge: one PStream failure-path gap symmetrical to the SIP adapter's, one test fixture that risks an unhandled 'error' throw. Low/Nit items optional.

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: Overcautious. error.message on a JS Error is already a controlled string from SIP.js's own code (e.g. "Connection timed out"), not a raw SDP dump. The SDP-fragment concern would require a very specific SIP.js code path, and even then the blast radius is "appears in customer logs" which is where every other error message in this SDK also ends up. Not worth truncating or adding doc-noise.

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: Tried this locally, since SipIceRestartConfig would be empty ({}), every object is assignable to it in TypeScript, so the union collapses to "any object" and doesn't actually discriminate, making the split no stronger than the current _config + docblock approach.

charliesantos (approved):

PR #436, opened 2026-05-26
[VBLOCKS-6609] sip reconnection policy

kpchoy:
**Contributing to Twilio**

## Pull Request Details

https://twilio-engineering.atlassian.net/browse/VBLOCKS-6609

### Description

Add preferred and primary backoff

### Before review
* [ ] Ready for review

### DIFF
--- a/lib/twilio/call.ts
+++ b/lib/twilio/call.ts
@@ -617,6 +617,7 @@ class Call extends EventEmitter {
     this._signalingAdapter.on('transportClose', this._onTransportClose);
     this._signalingAdapter.on('connected', this._onConnected);
     this._signalingAdapter.on('message', this._onMessageReceived);
+    this._signalingAdapter.on('iceRestartNeeded', this._onIceRestartNeeded);
 
     this.on('error', error => {
       this._publisher.error('connection', 'error', {
@@ -1152,6 +1153,7 @@ class Call extends EventEmitter {
       this._signalingAdapter.removeListener('transportClose', this._onTransportClose);
       this._signalingAdapter.removeListener('connected', this._onConnected);
       this._signalingAdapter.removeListener('message', this._onMessageReceived);
+      this._signalingAdapter.removeListener('iceRestartNeeded', this._onIceRestartNeeded);
     };
 
     // This is kind of a hack, but it lets us avoid rewriting more code.
@@ -1357,6 +1359,21 @@ class Call extends EventEmitter {
   }
 
   /**
+   * Called when the SIP signaling adapter recovered the WS while an
+   * ICE-restart re-INVITE was orphaned. The remote never received the new
+   * ICE/DTLS creds, so kick a fresh ICE restart.
+   */
+  private _onIceRestartNeeded = (payload: { callsid?: string }): void => {
+    if (payload?.callsid !== this.parameters.CallSid
+        || this.status() === Call.State.Closed) {
+      return;
+    }
+    this._log.info('iceRestartNeeded after WS recovery; retrying ICE restart');
+    this._mediaReconnectStartTime = Date.now();
+    this._onMediaFailure(Call.MediaFailure.ConnectionFailed);
+  }
+
+  /**
    * Called when the {@link Call} is hung up.
    * @param payload
    */
@@ -1425,6 +1442,14 @@ class Call extends EventEmitter {
 
       // This is a retry. Previous ICE Restart failed
       if (isEndOfIceCycle) {
+
+        // While signaling is down, an ICE restart re-INVITE cannot reach the
+        // server, so don't burn the budget or retry. iceRestartNeeded from the
+        // adapter will resume the loop with a fresh budget once WS recovers.
+        if (this._signalingAdapter.status === 'disconnected') {
+          this._log.info('Skipping ICE restart retry while signaling is disconnected');
+          return;
+        }
 
         // We already exceeded max retry time.
         if (Date.now() - this._mediaReconnectStartTime > BACKOFF_CONFIG.max) {
@@ -1654,7 +1679,14 @@ class Call extends EventEmitter {
     this._log.error('Received transportClose from pstream');
     this._log.debug('#transportClose');
     this.emit('transportClose');
-    if (this._signalingReconnectToken) {
+    // Stay alive while the call is in-progress. The transport may recover and
+    // (for SIP) the adapter will tell us to re-trigger ICE restart via
+    // 'iceRestartNeeded'. If recovery doesn't happen in time, media-reconnect
+    // backoff will eventually close the call. We only mark Closed up-front
+    // when the call hasn't been answered yet.
+    if (this._status === Call.State.Open
+        || this._status === Call.State.Reconnecting
+        || this._signalingReconnectToken) {
       this._status = Call.State.Reconnecting;
       this._signalingStatus = Call.State.Reconnecting;
       this._publisher.info('connection', 'reconnecting', null, this);
--- a/lib/twilio/call.ts
+++ b/lib/twilio/call.ts
@@ -1364,8 +1364,10 @@ class Call extends EventEmitter {
    * ICE/DTLS creds, so kick a fresh ICE restart.
    */
   private _onIceRestartNeeded = (payload: { callsid?: string }): void => {
-    if (payload?.callsid !== this.parameters.CallSid
-        || this.status() === Call.State.Closed) {
+    if (payload?.callsid !== this.parameters.CallSid) {
+      return;
+    }
+    if (this.status() !== Call.State.Open && this.status() !== Call.State.Reconnecting) {
       return;
     }
     this._log.info('iceRestartNeeded after WS recovery; retrying ICE restart');
--- a/lib/twilio/signaling/sipsignalingadapter.ts
+++ b/lib/twilio/signaling/sipsignalingadapter.ts
@@ -87,6 +87,16 @@ interface SdhBinding {
 }
 
 /**
+ * Per-invocation handle for an in-flight ICE-restart re-INVITE. Captured by
+ * the requestDelegate closures so onAccept/onReject can act on the specific
+ * invite that produced them, even after the adapter has moved on.
+ */
+interface InflightIceRestart {
+  callSid: string;
+  stale: boolean;
+}
+
+/**
  * Fallback SDH returned from the factory when no binding is registered
  * for a session. Every operation rejects with a clear error so SIP.js's
  * Promise-based accept/invite paths fail cleanly instead of crashing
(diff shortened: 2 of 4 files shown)


lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -700,21 +792,147 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
       });
       this._registerer = null;
     }
 
-    if (this._status !== 'disconnected') {
-      this._status = 'disconnected';
-      this.emit('transportClose');
-      this.emit('offline', this);
+    this._status = 'disconnected';
+    this.emit('transportClose');
```
charliesantos, line 831: `offline` is never emitted under the new code path — not on backoff exhaustion, not on `destroy()`, nowhere. Any consumer relying on `offline` (e.g. `Device`) to detect permanent signaling loss will never fire. Emit `offline` when the primary backoff budget is exhausted (and again on `destroy()` mid-reconnect), or explicitly document that this adapter never emits it and audit all consumers.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -30,6 +32,17 @@ import {
...
+// Reconnect backoff policy mirrors WSTransport's two-tier model:
+// - Preferred: short retries on the same URI for a bounded window (15s).
+// - Primary: unbounded fallback retries with a higher delay cap.
+// Twilio's SIP-over-WS exposes a single URI today, so primary retries the
+// same URI. Multi-edge SIP is on the roadmap; when it lands, primary is
+// the tier that should rotate through fallback edges (see WSTransport).
+const PREFERRED_BACKOFF_CONFIG = { factor: 2.0, jitter: 0.40, min: 100, max: 1000 };
+const PRIMARY_BACKOFF_CONFIG = { factor: 2.0, jitter: 0.40, min: 100, max: 20000 };
+const MAX_PREFERRED_DURATION_MS = 15000;
+const MAX_PRIMARY_DURATION_MS = Infinity;
```
charliesantos, line 44: `MAX_PRIMARY_DURATION_MS = Infinity` makes the expiry check in `_onPrimaryBackoffReady` (line 878) permanently unreachable, so the primary-tier termination path is dead code and untestable. Combined with the missing `offline` emit, a permanently-severed network leaves the call hanging forever. Define a finite cap and emit `offline` on expiry to match WSTransport, or remove the guard entirely with a TODO acknowledging infinite retries are intentional.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -700,21 +792,147 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+    if (!this._userAgent) {
+      return;
+    }
+    this._log.info('Reconnect succeeded');
+    this._isReconnecting = false;
+    this._resetBackoffs();
+
+    // SIP.js's onConnect delegate normally fires from reconnect() and runs
+    // _onTransportConnect. Defensive sync if it didn't.
+    if (this._status === 'disconnected') {
```
charliesantos: If SIP.js fires `onConnect` asynchronously after `reconnect()` resolves (e.g. as a microtask), this defensive branch sees `_status === 'disconnected'` and emits `connected`, then `onConnect`'s delegate emits `connected` a second time. Set `_status = 'connected'` before this check so a subsequent `onConnect` becomes a no-op, or guard `_onTransportConnect` itself to no-op when already connected. Add a test pinning `connected` fires exactly once.

lib/twilio/call.ts
```diff
@@ -1426,6 +1443,14 @@ class Call extends EventEmitter {
       // This is a retry. Previous ICE Restart failed
       if (isEndOfIceCycle) {
 
+        // While signaling is down, an ICE restart re-INVITE cannot reach the
+        // server, so don't burn the budget or retry. iceRestartNeeded from the
+        // adapter will resume the loop with a fresh budget once WS recovers.
+        if (this._signalingAdapter.status === 'disconnected') {
```
charliesantos, line 1451: This guard pauses the ICE retry budget while WS is down, but if `SipSignalingAdapter` retries forever (per H1/M1) and the network stays down, `iceRestartNeeded` never arrives and the call sits in `Reconnecting` indefinitely with no terminal signal. Add a wall-clock watchdog that closes the call if signaling hasn't recovered within a defined maximum, independent of the WS backoff tier.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -700,21 +792,147 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+    this._pendingIceRestartRecovery.forEach((callSid: string) => recovery.push(callSid));
+    this._pendingIceRestartRecovery.clear();
+    this._log.info(`_onReconnectSuccess: ${recovery.length} stale ICE-restart(s) to re-trigger`);
+    recovery.forEach((callSid: string) => {
+      this._log.info(`Emitting iceRestartNeeded for ${callSid}`);
+      this.emit('iceRestartNeeded', { callsid: callSid });
+    });
+  }
+
+  private _resetBackoffs(): void {
```
charliesantos, line 934: `_resetBackoffs` calls `backoff.reset()` and nulls `_backoffStartTime`, but reuses the same `Backoff` instances across cycles. The preferred 15s budget clock relies on `attempt === 0` re-firing the `backoff` listener to re-stamp `_backoffStartTime.preferred` — correct but subtle, and currently untested. Add a test that a second disconnect after a successful reconnect gets the full fresh 15s window.

lib/twilio/call.ts
```diff
@@ -1356,6 +1358,21 @@ class Call extends EventEmitter {
     }
   }
 
+  /**
+   * Called when the SIP signaling adapter recovered the WS while an
+   * ICE-restart re-INVITE was orphaned. The remote never received the new
+   * ICE/DTLS creds, so kick a fresh ICE restart.
+   */
+  private _onIceRestartNeeded = (payload: { callsid?: string }): void => {
+    if (payload?.callsid !== this.parameters.CallSid
```
charliesantos: The guard accepts any state except `Closed`, so an unexpected `iceRestartNeeded` during `Pending`/`Ringing` would call `_onMediaFailure(ConnectionFailed)` in an inappropriate state. Worth tightening to `this.status() === Call.State.Open || this.status() === Call.State.Reconnecting`.

tests/unit/sipsignalingadapter.ts
```diff
@@ -745,6 +752,19 @@ describe('SipSignalingAdapter', () => {
...
+
+    it('does not emit hangup when session.invite rejects with RequestPendingError', async () => {
+      const { adapter, inviterStub } = createAdapter();
+      adapter.invite('call-1', { sdp: 'sdp', params: 'To=bob', peerConnection: createPeerConnectionStub() });
+      inviterStub.state = 'Established';
+      inviterStub.invite = sinon.stub().returns(Promise.reject(new RequestPendingError('Reinvite in progress')));
+      const onHangup = sinon.stub();
+      adapter.on('hangup', onHangup);
+      adapter.iceRestart('call-1', { mediaHandler: mediaHandlerStub() });
+      await new Promise(resolve => setTimeout(resolve, 0));
```
charliesantos: `await new Promise(resolve => setTimeout(resolve, 0))` won't flush the promise microtask chain if this `describe` block ever gains `sinon.useFakeTimers()` — the assertion would trivially pass without the implementation running. Worth awaiting the actual promise chain, or switching to `clock.tickAsync(0)` if fake timers are introduced.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -403,6 +485,16 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
       },
     };
     session.invite(inviteOptions).catch((error: Error) => {
+      // A previous re-INVITE is still in flight — its requestDelegate owns the
+      // outcome. Treating this as fatal here would tear down the call before
+      // the in-flight re-INVITE has a chance to succeed or surface the real
+      // failure (timeout, reject, etc.).
+      if (error instanceof RequestPendingError) {
```
charliesantos, line 493: `_inFlightIceRestarts.delete(inflight)` runs in the `RequestPendingError` branch and again in the fallthrough — safe (`Set.delete` on a missing key is a no-op) but confusing. Worth hoisting the `delete` above the `if` so it runs once unconditionally.

tests/unit/call.ts
```diff
@@ -2790,6 +2790,93 @@ describe('Call', function() {
...
+
+      it('triggers a fresh ICE restart for the matching callSid', () => {
+        // Open state so _onMediaFailure goes through the entry-into-Reconnecting
+        // path that issues a backoff.
+        conn['_mediaStatus'] = Call.State.Open;
+        conn['_mediaReconnectBackoff'].backoff = sinon.stub();
+        conn['_mediaReconnectBackoff'].reset = sinon.stub();
+        mediaHandler.version.pc.iceConnectionState = 'failed';
+
+        pstream.emit('iceRestartNeeded', { callsid: 'CAtest' });
```
charliesantos: This test and the one below both verify `iceRestartNeeded` triggers a backoff; only the second checks `_mediaReconnectStartTime` reset (the first doesn't pre-set an expired start time, so it can't actually verify budget renewal). Worth merging the two or tightening the first to cover the budget-reset case.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -700,21 +792,147 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+      return;
+    }
+    this._log.info('Reconnect succeeded');
+    this._isReconnecting = false;
+    this._resetBackoffs();
+
+    // SIP.js's onConnect delegate normally fires from reconnect() and runs
+    // _onTransportConnect. Defensive sync if it didn't.
+    if (this._status === 'disconnected') {
+      this._onTransportConnect();
```
charliesantos: Question: When `_onReconnectSuccess` defensively calls `_onTransportConnect()` and SIP.js *also* fires `onConnect`, does `_onTransportConnect` guard against running twice in the same cycle (status already `'connected'`)? If not, `'connected'` fires twice and any listener that creates a registerer or resets state will double-execute.

charliesantos (changes requested):
Reconnect lifecycle is well-tested for the happy path, but the primary-tier termination path is currently unreachable: `MAX_PRIMARY_DURATION_MS = Infinity` makes the expiry guard dead code, and there is no `offline` emit anywhere on terminal failure — a permanently-broken WS leaves the call hung in `Reconnecting` indefinitely.

kpchoy:
@charliesantos

This mirrors `WSTransport`'s established policy. See `lib/twilio/wstransport.ts:11-12`:

  ```ts
  const MAX_PREFERRED_DURATION = 15000;
  const MAX_PRIMARY_DURATION = Infinity;
  ```

and the matching primary-`ready` guard at `:656` which is also unreachable under defaults. WSTransport doesn't emit `offline` on exhaustion either — only `'close'` on explicit `close()`.

  So infinite primary retries with no terminal signal is the existing behavior for PStream-based calls today; the SIP adapter is intentionally matching it. The `MAX_PRIMARY_DURATION_MS` constant is in place as the natural extension point
  if we ever want to make it overridable (as WSTransport does via options), but defaulting to `Infinity` keeps SIP and PStream calls behaviorally identical.

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: `MAX_PRIMARY_DURATION_MS = Infinity` mirrors WSTransport

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: https://github.com/twilio/twilio-voice.js/pull/436#issuecomment-4548536167

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: Good catch fixed. `_onTransportConnect` is now idempotent (early-returns when `_status === 'connected'`), so whichever path fires first wins and the second is a no-op. `_onReconnectSuccess` no longer needs the `_status === 'disconnected'` guard around its defensive call.

lib/twilio/call.ts
kpchoy: Same concern as #1/#2 on the Call side. PStream-based calls have the identical behavior today (WSTransport retries primary forever with no Call-side watchdog), so adding one only here would diverge SIP from PStream.

charliesantos:
The WSTransport parity holds, but Device doesn't subscribe to WSTransport - it subscribes to PStream, which sits on top and does emit offline on every transport close and on destroy, and that's what flips Device.State to Unregistered and sets _shouldReRegister via _onSignalingOffline. The new adapter only emits offline during initial-registration failures. after a successful register, a permanently-severed WS loops the primary backoff forever and Device stays Registered indefinitely. so SIP-mode calls don't get the same terminal signal PStream-mode calls do, and that's the real gap. How about emitting offline from _onTransportDisconnect (mirroring PStream's behavior, not WSTransport's) so Device and Call observe the same terminal state in both modes. the MAX_PRIMARY_DURATION_MS = Infinity default can stay.

kpchoy:
Done. Emit `offline` from `_onTransportDisconnect`. Verified end-to-end with WS-drop during an active call: Device flips Unregistered + `_shouldReRegister`, primary backoff recovers WS, `_onSignalingConnected` triggers re-register,
  REGISTER short-circuits via the existing-registerer guard (no double REGISTER on the wire), Device flips back to Registered. Tests updated to pin the new contract.

charliesantos (approved):
Thanks for the offline emit. Clean PStream parity. LGTM.

PR #437, opened 2026-05-28
[VBLOCKS-6610] deferred ice restart

kpchoy:
**Contributing to Twilio**

## Pull Request Details

https://twilio-engineering.atlassian.net/browse/VBLOCKS-6610

### Description

Added a defer guard in SipSignalingAdapter.iceRestart() that skips the re-INVITE when _isReconnecting is true and queues the callSid in the existing recovery set, with 4 new unit tests covering the deferred path.

### Before review
* [ ] Ready for review

### DIFF
--- a/lib/twilio/signaling/sipsignalingadapter.ts
+++ b/lib/twilio/signaling/sipsignalingadapter.ts
@@ -403,6 +403,10 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
   }
 
   hangup(callSid: string, _config: HangupConfig = {}): void {
+    // Don't let _onReconnectSuccess emit 'iceRestartNeeded' for a call hung up
+    // mid-reconnect.
+    this._pendingIceRestartRecovery.delete(callSid);
+
     const pendingInvitation = this._pendingInvitations.get(callSid);
     if (pendingInvitation) {
       return this._hangupPendingInvitation(pendingInvitation, callSid);
@@ -441,7 +445,9 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
    * is unused here (the PStream adapter needs it; we preserve the shared
    * interface). All failure paths (no session, onReject, catch) emit
    * 'hangup' for the callSid so Call can clean up its inline re-INVITE
-   * listeners and transition out of the reconnecting state.
+   * listeners and transition out of the reconnecting state. The exception is
+   * when the transport is reconnecting: instead of emitting 'hangup' we queue
+   * the callSid for an 'iceRestartNeeded' emit on reconnect.
    */
   iceRestart(callSid: string, _config: IceRestartConfig): void {
     const session = this._getSession(callSid);
--- a/lib/twilio/signaling/sipsignalingadapter.ts
+++ b/lib/twilio/signaling/sipsignalingadapter.ts
@@ -87,6 +87,16 @@ interface SdhBinding {
 }
 
 /**
+ * Per-invocation handle for an in-flight ICE-restart re-INVITE. Captured by
+ * the requestDelegate closures so onAccept/onReject can act on the specific
+ * invite that produced them, even after the adapter has moved on.
+ */
+interface InflightIceRestart {
+  callSid: string;
+  stale: boolean;
+}
+
+/**
  * Fallback SDH returned from the factory when no binding is registered
  * for a session. Every operation rejects with a clear error so SIP.js's
  * Promise-based accept/invite paths fail cleanly instead of crashing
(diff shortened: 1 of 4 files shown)


lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -450,6 +450,14 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
       this.emit('hangup', { callsid: callSid });
       return;
     }
+    // Sending session.invite() over a dead transport queues a doomed re-INVITE
+    // that will 408 once SIP.js Timer B expires (~32s). Record the intent and
+    // let _onReconnectSuccess emit 'iceRestartNeeded' once the WS is back.
+    if (this._isReconnecting) {
+      this._log.info(`iceRestart deferred while reconnecting: ${callSid}`);
+      this._pendingIceRestartRecovery.add(callSid);
```
charliesantos, line 464: Hangup during reconnect doesn't clear `_pendingIceRestartRecovery`. If `Call` invokes `hangup(callSid)` after this branch queues the callSid, `_onReconnectSuccess` will still emit `iceRestartNeeded` for the now-dead callSid (and `_onTransportDisconnect` at line 824 has the same gap pre-existing). Worth adding `this._pendingIceRestartRecovery.delete(callSid)` in `hangup()` (or in both `_hangup*Session` helpers) to close both leaks in one line.

tests/unit/sipsignalingadapter.ts
```diff
@@ -1282,6 +1282,78 @@ describe('SipSignalingAdapter', () => {
...
+      adapter.invite('call-1', { sdp: 'sdp', params: 'To=bob', peerConnection: createPeerConnectionStub() });
+      inviterStub.state = 'Established';
+      uaStub._triggerConnect();
+      // Reset to count only iceRestart-driven invite() calls below.
+      inviterStub.invite.resetHistory();
+
+      uaStub._triggerDisconnect();
+      adapter.iceRestart('call-1', { mediaHandler: { iceRestart: sinon.stub() } });
+
+      sinon.assert.notCalled(inviterStub.invite);
```
charliesantos, line 1296: This test asserts only that no invite was sent — a regression that omitted the `.add(callSid)` call would still pass here (the next test would catch it via the round-trip assertion, but a direct positive check fails sharper). Worth adding `assert.ok((adapter as any)._pendingIceRestartRecovery.has('call-1'));` after the `notCalled` assertion.

tests/unit/sipsignalingadapter.ts
```diff
@@ -1282,6 +1282,78 @@ describe('SipSignalingAdapter', () => {
...
+      adapter.iceRestart('call-1', { mediaHandler: { iceRestart: sinon.stub() } });
+
+      uaStub._setReconnectImpl(() => { uaStub._triggerConnect(); return Promise.resolve(); });
+      await clock.tickAsync(100);
+
+      sinon.assert.calledOnce(onIceRestartNeeded);
+    });
+
+    it('emits iceRestartNeeded for each distinct callSid deferred during the outage', async () => {
+      const { adapter, uaStub, inviterStub } = createAdapter();
```
charliesantos, line 1339: This test relies on `createAdapter()`'s shared `inviterStub` covering both `'call-1'` and `'call-2'` sessions when `state` is set once. Worth a one-line comment, or setting state per-session, so a future per-call stub refactor doesn't break the setup silently.

lib/twilio/signaling/sipsignalingadapter.ts
charliesantos, line 464: Question: this is a behavioral divergence from `PStreamSignalingAdapter.iceRestart`, which calls `mediaHandler.iceRestart(...)` and ships the reinvite synchronously regardless of transport state. After this PR, the SIP path silently defers (no PC ICE gathering, no wire activity) and waits for `iceRestartNeeded` to round-trip back through `Call`. Is `Call` aware of that asymmetry, or does it currently assume both adapters drive the re-INVITE immediately?

charliesantos (changes requested):
Small change, looks correct and well-tested for the happy paths; a few minor follow-ups inline. One docblock note: the `iceRestart()` JSDoc at lines 437–445 still says "all failure paths emit hangup", which is no longer true now that the new `_isReconnecting` branch returns without emitting — worth appending one line about the queue-on-reconnect behavior.

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: Yes — `Call` handles it via the `iceRestartNeeded` round-trip. `Call._onIceRestartNeeded` (`call.ts:1366`) re-drives `_onMediaFailure(ConnectionFailed)` once the WS recovers, and during the outage `Call` is already in the `Reconnecting` media-reconnect flow, so the gap is covered by the existing media-reconnect timeout rather than a wire re-INVITE.

  PStream doesn't defer because its transport handles reconnection itself; the SIP path can't, since a re-INVITE on a dead SIP.js transport just 408s ~32s later (Timer B). So `Call` doesn't assume immediate re-INVITE from both adapters.

charliesantos (approved):

PR #441, opened 2026-07-07
[VBLOCKS-6900] feat: build ci

kpchoy:
**Contributing to Twilio**

## Pull Request Details

https://twilio-engineering.atlassian.net/browse/VBLOCKS-6900

### Description

GHA build CI

### Before review
* [ ] Ready for review

### DIFF
--- a/.github/actions/artifactory-oidc/action.yml
+++ b/.github/actions/artifactory-oidc/action.yml
@@ -1,0 +1,63 @@
+name: 'Artifactory OIDC Auth'
+description: 'Exchange GitHub OIDC token for Artifactory access token and configure npm'
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Exchange OIDC token for Artifactory access token
+      shell: bash
+      run: |
+        set -euo pipefail
+
+        # jq is used throughout; fail early with a clear message if it's absent
+        # (present on ubuntu-latest today, but not guaranteed on every runner).
+        if ! command -v jq >/dev/null 2>&1; then
+          echo "::error::jq is required by the artifactory-oidc action but was not found on the runner."
+          exit 1
+        fi
+
+        # Normalize the base URL: strip any trailing slashes so REGISTRY and the
+        # //host/... auth line never end up with double slashes.
+        BASE_URL=$(echo "${ARTIFACTORY_URL}" | sed -E 's#/+$##')
+        if [ -z "$BASE_URL" ]; then
+          echo "::error::ARTIFACTORY_URL is empty."
+          exit 1
+        fi
+        # HOST is scheme-stripped and path-stripped (everything before the first /).
+        HOST=$(echo "${BASE_URL}" | sed -E 's#^https?://##; s#/.*$##')
+
+        # 1. Request a GitHub OIDC JWT with Artifactory as the audience.
+        GITHUB_JWT=$(curl -sS \
+          -H "Authorization: bearer ${ACTIONS_ID_TOKEN_REQUEST_TOKEN}" \
+          "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=${BASE_URL}" \
+          | jq -r '.value')
+        echo "::add-mask::$GITHUB_JWT"
+
+        # 2. Exchange it for an Artifactory access token. The body MUST be JSON
+        #    and MUST include provider_name, or JFrog issues no token.
+        RESP=$(curl -sS "${BASE_URL}/access/api/v1/oidc/token" \
+          -H 'Content-Type: application/json' \
+          -d "{\"grant_type\":\"urn:ietf:params:oauth:grant-type:token-exchange\",
+               \"subject_token_type\":\"urn:ietf:params:oauth:token-type:id_token\",
+               \"subject_token\":\"${GITHUB_JWT}\",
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,139 @@
...
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: ${{ matrix.node }}
+          cache: npm
+      - run: npm ci --ignore-scripts
+      # Build first: test:es5/test:esm validate the es5/ and esm/ build outputs.
+      - run: npm run build
+      - run: npm run test:es5
+      - run: npm run test:esm
+      - run: npm run test:unit
```
charliesantos, line 110: Node 24 matrix leg will fail here: the runtime's experimental TypeScript stripping conflicts with ts-node, so `npm run test:unit` (nyc mocha -r ts-node/register) crashes on import. Set `NODE_OPTIONS=--no-experimental-strip-types` on this step (an env block, or a matrix include scoped to node 24).

.github/actions/artifactory-oidc/action.yml
```diff
@@ -0,0 +1,63 @@
...
+        BASE_URL=$(echo "${ARTIFACTORY_URL}" | sed -E 's#/+$##')
+        if [ -z "$BASE_URL" ]; then
+          echo "::error::ARTIFACTORY_URL is empty."
+          exit 1
+        fi
+        # HOST is scheme-stripped and path-stripped (everything before the first /).
+        HOST=$(echo "${BASE_URL}" | sed -E 's#^https?://##; s#/.*$##')
+
+        # 1. Request a GitHub OIDC JWT with Artifactory as the audience.
+        GITHUB_JWT=$(curl -sS \
```
charliesantos: Add `--fail` to this OIDC JWT fetch (`curl -sSf`). On a 401/403/500 curl still exits 0, jq yields the literal string `null`, GITHUB_JWT becomes "null", `::add-mask::` is a no-op, and it gets forwarded as subject_token - the real failure then surfaces later as a misleading "OIDC token exchange failed". Failing fast here reports the true root cause.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,139 @@
...
+
+on:
+  pull_request:
+  push:
+    branches: [master]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  id-token: write
```
charliesantos: `id-token: write` at workflow level grants OIDC-minting to every job, including all-checks which has no OIDC need. Worth scoping it into each job's `permissions:` block (lint/typecheck/test/build) and leaving `contents: read` at the top to limit blast radius.

.github/actions/artifactory-oidc/action.yml
```diff
@@ -0,0 +1,63 @@
+name: 'Artifactory OIDC Auth'
```
charliesantos, line 1: The action reads `ARTIFACTORY_URL` from the ambient environment rather than declaring it as an input, so a reader can't tell what it requires without the calling workflow, and actionlint can't flag a missing var. Worth declaring `inputs.artifactory_url` (required) and referencing `${{ inputs.artifactory_url }}`, with callers passing it via `with:`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,139 @@
+# Platform build/CI for @twilio/voice-sdk (VBLOCKS-6900, ADR 1634 / PEA-198).
+# Modeled on twilio/twilio-agent-connect-typescript's ci.yml, adapted for voice-js:
+#   - NO fork CI (lead decision 2026-07-06, stricter than the program). Jobs run
+#     unconditionally on the trusted runner; the job-level `if` guard is a
+#     defense-in-depth backstop so a fork PR can never execute even if repo
+#     settings drift. See VBLOCKS-6900 and the no-fork-CI decision.
+#   - Every `uses:` is full-SHA-pinned (repo enforces sha_pinning_required: true,
+#     allowed_actions: selected; tags are rejected at startup).
+#   - e2e (Cypress/docker) is out of scope here — owned by VBLOCKS-6876.
+name: CI
```
charliesantos, line 10: Question: Should we remove circleci related configs/code?

charliesantos (changes requested):
CI infra change looks solid - SHA-pinned actions, correct token masking, and no injection vectors. Requesting changes on one blocker: the Node 24 test leg fails deterministically without NODE_OPTIONS=--no-experimental-strip-types.

.github/workflows/ci.yml
kpchoy: The test for node 24 is green. `ts-node/register's` require hook takes precedence over the runtime's TS stripping, so the conflict doesn't trigger. Holding off on `--no-experimental-strip-types` since it's green on evidence, happy to add it if a real failure shows up.

.github/workflows/ci.yml
kpchoy: Commented out the circleci PR workflow for cleaner PRs. Will keep circleci code for reference for now. Will remove when gha migration complete

charliesantos (approved):

PR #450, opened 2026-07-24
feat: implement OIDC token minting for e2e tests

luisrivas:
## Summary
Adds `.github/workflows/e2e.yml`, running the integration suite (`test:integration`) in GitHub Actions across Chrome and Firefox and each browser's stable/beta/unstable channels.

`tests/lib/token.js` is refactored to route through a new `tests/lib/vendor.js` client so the same test code works unchanged in the new workflow. Long-lived API key secrets are removed from the Cypress environment; the workflow authenticates with a GitHub Actions OIDC token instead.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Start Xvfb and dbus
         shell: bash
         run: |
-          set -u
+          set -euo pipefail
           sudo dbus-daemon --system --fork
           Xvfb $DISPLAY -screen 0 1280x1024x24 &
           xvfb_pid=$!
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,72 @@
+name: E2E
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}
+
+jobs:
+  e2e:
+    name: e2e (${{ matrix.browser }}, ${{ matrix.bver }})
+    if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
+    runs-on: ubuntu-latest-large
+    container:
+      image: twilio/twilio-video-browsers:${{ matrix.browser }}-${{ matrix.bver }}
+    env:
+      DISPLAY: ':99'
+    permissions:
+      contents: read
+      id-token: write
+    strategy:
+      fail-fast: false
+      matrix:
+        browser: [chrome, firefox]
+        bver: [stable, beta, unstable]
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      - name: Artifactory OIDC Auth
+        uses: ./.github/actions/artifactory-oidc
+        with:
+          artifactory_url: ${{ vars.ARTIFACTORY_URL }}
+
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 22
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci --ignore-scripts
+
+      - name: Install Cypress binary
+        run: ./node_modules/.bin/cypress install
+
+      # Container jobs skip the image's own ENTRYPOINT that normally starts these.
+      - name: Start Xvfb and dbus
+        shell: bash
+        run: |
+          sudo dbus-daemon --system --fork
+          Xvfb $DISPLAY -screen 0 1280x1024x24 &
+          sleep 1
+
+      - name: Run e2e integration tests (chrome)
+        if: matrix.browser == 'chrome'
+        env:
+          VENDOR_URL: ${{ vars.VENDOR_URL }}
+          VENDOR_AUDIENCE: ${{ vars.VENDOR_URL }}
+        run: npm run test:integration:chrome
+
+      - name: Run e2e integration tests (firefox)
+        if: matrix.browser == 'firefox'
+        env:
+          VENDOR_URL: ${{ vars.VENDOR_URL }}
+          VENDOR_AUDIENCE: ${{ vars.VENDOR_URL }}
+        run: npm run test:integration:firefox
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -68,9 +68,10 @@ jobs:
           VENDOR_AUDIENCE: ${{ vars.VENDOR_URL }}
         run: npm run test:integration:chrome
 
+      # Cypress can't auto-detect Firefox's beta/nightly builds by name.
       - name: Run e2e integration tests (firefox)
         if: matrix.browser == 'firefox'
         env:
           VENDOR_URL: ${{ vars.VENDOR_URL }}
           VENDOR_AUDIENCE: ${{ vars.VENDOR_URL }}
-        run: npm run test:integration:firefox
+        run: npm run test:integration -- --browser /usr/local/bin/firefox
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -49,6 +49,10 @@ jobs:
       - name: Install Cypress binary
         run: ./node_modules/.bin/cypress install
 
+      # lib/twilio/constants.ts is gitignored and generated at build time.
+      - name: Build constants
+        run: npm run build:constants
+
       # Container jobs skip the image's own ENTRYPOINT that normally starts these.
       - name: Start Xvfb and dbus
         shell: bash
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -57,9 +57,17 @@ jobs:
       - name: Start Xvfb and dbus
         shell: bash
         run: |
+          set -u
           sudo dbus-daemon --system --fork
           Xvfb $DISPLAY -screen 0 1280x1024x24 &
-          sleep 1
+          xvfb_pid=$!
+          socket="/tmp/.X11-unix/X${DISPLAY#:}"
+          for _ in $(seq 1 50); do
+            [ -e "$socket" ] && break
+            kill -0 "$xvfb_pid" 2>/dev/null || { echo "Xvfb exited during startup"; exit 1; }
+            sleep 0.2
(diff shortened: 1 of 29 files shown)


tests/lib/vendorProxy.js
```diff
@@ -0,0 +1,158 @@
...
+  stop() {
+    if (this._server) {
+      this._server.close();
+      this._server = null;
+    }
+  }
+
+  async _handleRequest(req, res) {
+    // This port is a different origin than the page Cypress serves.
+    res.setHeader('Access-Control-Allow-Origin', '*');
```
kpchoy, line 62: **[High] Vending proxy is a confused deputy: open CORS + no inbound auth**

The proxy holds a privileged OIDC token and, for any POST to `/vend`, attaches it server-side and forwards the caller's body to the vendor, returning the response verbatim. There's no inbound auth or `action` allow-list, and CORS is `*`, so a browser can send the request cross-origin and read the minted Twilio voice token. Any co-process on the runner (or, in local runs, any origin loaded in the same browser that reaches the loopback port) can mint call-capable credentials, the class of secret this PR set out to protect.

_In ephemeral CI the practical blast radius is small (random loopback port, container torn down after); it's most exploitable in local runs. Still worth hardening since the proxy is now the single credential chokepoint._

**Suggestion:** generate a per-run secret in `start()`, inject it only into the Cypress test env, require it on `/vend`, and reflect a specific origin instead of `*`. Add an allow-list of permitted `action` values.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,77 @@
...
+      # lib/twilio/constants.ts is gitignored and generated at build time.
+      - name: Build constants
+        run: npm run build:constants
+
+      # Container jobs skip the image's own ENTRYPOINT that normally starts these.
+      - name: Start Xvfb and dbus
+        shell: bash
+        run: |
+          sudo dbus-daemon --system --fork
+          Xvfb $DISPLAY -screen 0 1280x1024x24 &
```
kpchoy, line 62: **[High] Xvfb backgrounded + fixed `sleep 1` hides startup failure**

`Xvfb $DISPLAY ... &` then `sleep 1` means the step exits 0 whether or not the display came up. If Xvfb dies (display already in use, package missing), the step stays green and the real failure surfaces minutes later inside Cypress as a confusing "cannot open display", turning CI red for the wrong reason.

**Suggestion:** `Xvfb ... & XVFB_PID=$!`, poll `xdpyinfo -display $DISPLAY` for readiness, then `kill -0 $XVFB_PID || { echo 'Xvfb failed'; exit 1; }`. Add `set -euo pipefail` to the step.

kpchoy (commented):
Thanks Luis, really nice work here. Clean migration off the long-lived key secrets, and the async token refactor is tidy across the whole suite. Left a few notes inline.

tests/lib/vendorProxy.js
```diff
@@ -0,0 +1,158 @@
...
+      res.end(JSON.stringify({ error: 'not found' }));
+      return;
+    }
+
+    try {
+      const body = await readBody(req);
+      const { status, text } = await this._forward(body);
+      res.writeHead(status, { 'Content-Type': 'application/json' });
+      res.end(text);
+    } catch (error) {
```
kpchoy, line 88: **[Medium] Proxy errors are invisible server-side and reflected to callers**

This catch returns `{error: error.message}` (carrying `Access-Control-Allow-Origin: *`) but logs nothing to the proxy's stdout/stderr. An OIDC/vendor failure yields a 500 the spec may discard, leaving zero diagnostic trail in CI while leaking the internal message cross-origin.

**Suggestion:** `console.error('[vendorProxy] /vend failed', error)` (with stack) before responding, and return a generic body (e.g. `{ error: 'internal error' }`) to the caller.

tests/lib/vendor.js
```diff
@@ -0,0 +1,37 @@
...
+  const response = await fetch(proxyUrl, {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify(Object.assign({ action }, params)),
+  });
+
+  const text = await response.text();
+  let parsed;
+  try {
+    parsed = JSON.parse(text);
```
kpchoy: **[Medium] `callVendor` parses JSON before the `res.ok` check**

A vendor 5xx returning an HTML/text error page throws `"non-JSON response from vendor: …"` instead of surfacing the real status; the 502 is mislabeled as a parse error. (The raw body text is included in the message, so it's degraded diagnostics, not total loss.)

**Suggestion:** check `res.ok` first, throw with status + raw text, then `JSON.parse` only on the success path.

.github/workflows/e2e.yml
charliesantos, line 62: Building on @kpchoy's point — `Xvfb ... &` then `sleep 1` exits 0 whether or not the display came up, so a startup failure only surfaces minutes later inside Cypress as "cannot open display", and the step has no `set -euo pipefail`. Add `set -euo pipefail`, capture the PID, and poll `xdpyinfo` for readiness, failing the step if Xvfb dies. The `dbus-daemon` line has the same exposure.

cypress/e2e/stirshaken.cy.ts
```diff
@@ -67,10 +69,10 @@ describe('SHAKEN/STIR', function() {
           return devShim._stream.transport.__onSocketMessage(message);
         };
 
-        call1 = await (device1['connect'] as any)({
-          params: { CallerId: (env as any).callerId },
-        });
-      }));
+        call1 = await (device1['connect'] as any)();
```
charliesantos, line 72: Dropping the explicit `CallerId` here relies on the server-side `stir` variant to trigger attestation. If it doesn't fire, the later `call2!.callerInfo!.isVerified` assertion throws a TypeError on null (the non-null assertions are compile-time only) that reads as a harness bug rather than "attestation didn't fire". Guard it: `assert.notEqual(call2!.callerInfo, null, 'callerInfo is null - STIR attestation did not fire')` before asserting `isVerified`.

tests/lib/vendorProxy.js
```diff
@@ -0,0 +1,158 @@
...
+      res.writeHead(500, { 'Content-Type': 'application/json' });
```
charliesantos, line 92: Building on @kpchoy's point — this catch reflects `error.message` cross-origin (with `Access-Control-Allow-Origin: *`) and logs nothing server-side, so an OIDC/upstream failure can leak the request URL or audience while leaving no trace in the CI log. Add `console.error('[vendorProxy] /vend error', error)` and return a generic `{ error: 'internal error' }` to callers.

tests/lib/vendor.js
charliesantos: Building on @kpchoy's point — `JSON.parse` runs before the `response.ok` check, so a 5xx with an HTML body throws a misleading parse error and discards the real status. Consider checking `response.ok` first and throwing with the status + raw text, then parsing JSON only on success.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,77 @@
...
+          sleep 1
+
+      - name: Run e2e integration tests (chrome)
+        if: matrix.browser == 'chrome'
+        env:
+          VENDOR_URL: ${{ vars.VENDOR_URL }}
+          VENDOR_AUDIENCE: ${{ vars.VENDOR_URL }}
```
charliesantos, line 76: Question: `VENDOR_AUDIENCE` is set to the endpoint URL, which the vendor validates against its own `GITHUB_OIDC_AUDIENCE` config. Can you confirm the deployed instance's `GITHUB_OIDC_AUDIENCE` is exactly this string? If the two drift (trailing slash, path change), OIDC verification fails closed with a 401 that's only diagnosable from the Function logs.

charliesantos (changes requested):
Nice migration off the long-lived key secrets, and the async token refactor is clean across the suite. A few code-level items remain: the Xvfb fail-open, a STIR null guard, and proxy error handling. Details inline.

tests/lib/vendorProxy.js
luisrivas: Fixed with a per-run secret.

.github/workflows/e2e.yml
luisrivas: Fixed, `set -euo pipefail` added. Used a socket poll, not `xdpyinfo`.

tests/lib/vendorProxy.js
luisrivas: Fixed. Logs the error, returns a generic body.

tests/lib/vendor.js
luisrivas: Fixed. Status checked first.

.github/workflows/e2e.yml
luisrivas: Fixed.

cypress/e2e/stirshaken.cy.ts
luisrivas: Fixed. Guard added.

tests/lib/vendor.js
luisrivas: Fixed

.github/workflows/e2e.yml
luisrivas: Confirmed, they match exactly. I reused `${{ vars.VENDOR_URL }}` to reduce friction, but the audience can be any agreed string that we match on the server side. I agree there's room to improve it. Any preference on your side?

cypress.config.ts
```diff
@@ -1,19 +1,14 @@
 const { defineConfig } = require('cypress');
+const VendorProxy = require('./tests/lib/vendorProxy');
 
 module.exports = defineConfig({
   env: {
-    ACCOUNT_SID: process.env.ACCOUNT_SID,
     AUTH_TOKEN: process.env.AUTH_TOKEN,
```
kpchoy, line 6: **[Nit] `AUTH_TOKEN` / `env.js` plumbing is now vestigial on the e2e path**

With minting routed through the vendor and `stirshaken.cy.ts` dropping its `env` import, nothing in the e2e specs reads `AUTH_TOKEN`, `accountSid`, or `appSid` anymore. `tests/env.js` still declares and "requires" all three, but the check passes on key presence regardless of value, so it isn't really guarding anything. The only remaining consumers (`generateCapabilityToken`, `tests/docker/proxy/server.js`) aren't exercised by this workflow. Not blocking and fine to defer, but if you want to close the loop on the secret removal, dropping the `AUTH_TOKEN` passthrough here and trimming `env.js` to what's actually used would make the migration self-contained.

kpchoy (approved):
looks good to me!

.github/workflows/e2e.yml
charliesantos: Thanks for checking. This is good to me for now.

charliesantos (approved):

PR #452, opened 2026-08-04
[VBLOCKS-6989] send digits playback

kpchoy:
**Contributing to Twilio**

## Pull Request Details

https://twilio-engineering.atlassian.net/browse/VBLOCKS-6989

### Description

- Fixed the w pause stacking on top of a run's tones instead of overlapping them, so post-pause playback now starts in step with the wire.
- Added the wire-vs-local unit test plus two Cypress regression tests (7w8, 45w6) and a CHANGELOG entry.
- Hardened those tests so they can actually fail, and fixed the CHANGELOG, which had the drift direction backwards.

### Before review
* [x] Ready for review


luisrivas (dismissed):
Thanks, @kpchoy! The `CHANGELOG` is promising a bit more than the code delivers, though. It reads as full sync, but a 500ms pause can only absorb two 200ms tones. Runs of three or more hit the clamp and the leftover accumulates: `123w456w789` comes out 100ms late, then 202ms late. Worth rewording so the entry matches the behavior, since this one is user-facing and people will read it as exact.

luisrivas (approved):
Thanks, @kpchoy !

PR #455, opened 2026-08-12
[VBLOCKS-7004] connection resilience

kpchoy:
**Contributing to Twilio**

## Pull Request Details

https://twilio-engineering.atlassian.net/browse/VBLOCKS-7004

### Description

- reconnect with backoff

### Before review
* [x] Ready for review

### DIFF
--- a/lib/twilio/signaling/sipclosecodehook.ts
+++ b/lib/twilio/signaling/sipclosecodehook.ts
@@ -1,50 +1,36 @@
 import Log from '../log';
 
 /**
- * Holds the close code of the most recent WebSocket close event, or
- * undefined if no close has been observed (or the hook failed to install).
+ * Close code of the most recent WebSocket close, or undefined if none has been
+ * observed (or the hook failed to install). The reader clears it once consumed.
  */
 export interface CloseCodeRecorder {
   lastCloseCode?: number;
 }
 
 /**
- * SIP.js's Transport does not expose the WebSocket close code. Its close
- * handler folds the code into a human-readable Error message
- * (`WebSocket closed <server> (code: 1006)`) and hands only that Error to
- * `UserAgentDelegate.onDisconnect`. Parsing that string would couple us to a
- * log format with no API guarantee.
+ * SIP.js only exposes the close code inside the Error message it hands to
+ * `onDisconnect`, so capture it structurally instead of parsing that string.
  *
- * Instead we wrap the transport's close handler. SIP.js registers its
- * listener as `ws.addEventListener('close', (ev) => this.onWebSocketClose(ev, ws))`,
- * so the dispatch goes through `this` and an own property on the instance
- * takes precedence over the prototype method. Ours therefore runs BEFORE the
- * base implementation, which is what lets us record the code before
- * `onDisconnect` fires.
+ * SIP.js dispatches through `this.onWebSocketClose(...)`, so an own property
+ * shadows the prototype method and runs before it, which is what lets us
+ * record the code before `onDisconnect` fires. A second `close` listener would
+ * run too late, and subclassing fails under this SDK's ES5 build.
  *
- * Two alternatives do not work:
- *  - Adding a second listener via the public `ws` getter: SIP.js registers
- *    its own listener when the socket is created, so ours would run after
- *    `onDisconnect` had already been dispatched.
- *  - Subclassing Transport: this SDK builds to ES5, and the emitted
- *    `__extends` helper cannot `super.call()` SIP.js's native ES6 class.
- *
- * SIP.js creates its Transport once in the UserAgent constructor and reuses
- * it across `reconnect()`, so a single install covers the whole lifetime.
+ * The Transport is created once and reused across `reconnect()`, so one
+ * install covers the whole lifetime.
  *
  * @param transport - the SIP.js Transport instance to wrap.
  * @param log - logger used to report a failed install.
- * @returns a recorder whose `lastCloseCode` is updated on every close.
+ * @returns a recorder updated on every close of the active socket.
  */
 export function installCloseCodeHook(transport: any, log: Log): CloseCodeRecorder {
   const recorder: CloseCodeRecorder = {};
 
   const baseOnWebSocketClose = transport?.onWebSocketClose;
 
-  // Degrade rather than throw. A missing method means SIP.js renamed or
-  // removed it in an upgrade; losing close codes costs a diagnostic, whereas
-  // throwing here would break calling entirely. The unit test asserting the
-  // method still exists is what catches this at build time.
+  // A SIP.js rename should cost a diagnostic, not break calling. The unit test
+  // asserting the method still exists is what catches the upgrade.
   if (typeof baseOnWebSocketClose !== 'function') {
     log.error(
       'Could not install WebSocket close-code hook: transport.onWebSocketClose ' +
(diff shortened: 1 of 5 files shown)


lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -803,6 +864,69 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+    this._log.info(`_onTransportConnect: ${recovery.length} stale ICE-restart(s) to re-trigger`);
+    recovery.forEach((callSid: string) => {
+      this._log.info(`Emitting iceRestartNeeded for ${callSid}`);
+      this.emit('iceRestartNeeded', { callsid: callSid });
+    });
+  }
+
+  /**
+   * Replay whatever queued while the transport was down. Drained before
+   * replaying, so a send that fails again re-queues rather than looping.
```
charliesantos: This comment is inaccurate: `_isReconnecting` is already cleared to false before the flush runs, so a failing `_sendInvite` does not re-queue - it just fails. Replace it with what `splice(0)` actually guarantees: draining atomically so a re-entrant flush from a send callback can't double-replay the queue.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -908,9 +1068,37 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
       return;
```
charliesantos, line 1072: `_attemptReconnect` bundles tier logging, the `settled` flag, the per-attempt timeout, the `reconnect()` promise, and backoff rescheduling on every path. The `settled` invariant is the safety contract but lives inline, so a future async path can miss it. Consider extracting a `_withConnectTimeout(tier, promise)` helper that owns `settled`, `clearTimeout`, and timeout-backoff scheduling, leaving this method linear.

charliesantos (commented):
Reviewed at head 99f6f582. The reconnection logic holds up on a close read: the second `_onTransportConnect()` on reconnect is a no-op via the `_status === 'connected'` guard, the message-queue flush is synchronous so `hangup()` can't race a session created mid-flush, and a `register()` issued during reconnect is captured by `_wasRegistered` before the defer guard. One stale JSDoc comment worth fixing plus a few minor cleanups below. Also worth updating the CHANGELOG - the change does more than 'reconnect with backoff' (close-code detection, INVITE queuing, connect-success timeout, per-attempt ceiling).

lib/twilio/signaling/sipclosecodehook.ts
```diff
@@ -0,0 +1,52 @@
...
+      'Could not install WebSocket close-code hook: transport.onWebSocketClose ' +
+      'is not a function. Close codes will be unavailable.',
+    );
+    return recorder;
+  }
+
+  transport.onWebSocketClose = (event: CloseEvent, ws: WebSocket): void => {
+    // Mirror SIP.js's own `if (ws !== this._ws) { return; }`: a replaced socket
+    // can still emit its close, and its code isn't the live connection's.
+    if (ws === transport.ws) {
```
charliesantos, line 45: This guard only holds if `transport.ws` is a stable property reference. A future sip.js upgrade that turns `ws` into a computed getter would make the check always fail and silently drop every close code. Worth adding a canary asserting `transport.ws` is a stable reference alongside the existing `onWebSocketClose` canary.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -343,6 +381,8 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
     this._pendingInvitations.clear();
```
charliesantos, line 387: The per-attempt `setTimeout(..., CONNECT_TIMEOUT_MS)` in `_attemptReconnect` is a local const, so it isn't cancelled here and still fires after destroy (harmless today via the null-safe `_backoff?.`, but a stray timer). Worth storing the handle and calling `clearTimeout` in destroy().

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -715,6 +756,22 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+    // Gated on _isReconnecting, not "not yet connected": Device awaits the
+    // connected promise before inviting, and queueing pre-connect would strand
+    // invites with no reconnect cycle to flush them.
+    if (this._isReconnecting) {
+      this._log.info(`Transport is down; queueing invite for ${tempCallSid}`);
+      this._messageQueue.push({
+        callSid: tempCallSid,
+        send: () => this._sendInvite(tempCallSid, config, reconnectToken),
+      });
+      this._emitTransportError();
```
charliesantos, line 777: `_emitTransportError()` fires synchronously in the invite queue path, so a caller that attaches its error handler after calling invite() would miss the 31009. Consider deferring via `Promise.resolve().then(...)` to match the async error-emission convention elsewhere.

tests/unit/sipsignalingadapter.ts
```diff
@@ -1205,6 +1328,117 @@ describe('SipSignalingAdapter', () => {
...
+      uaStub._triggerDisconnect();
+      await clock.tickAsync(5100);
+
+      const messages = warnSpy.getCalls().map((c: any) => String(c.args[0]));
+      assert(
+        messages.some((m: string) => m.includes('Reconnect attempt failed')),
+        'the underlying connect error must reach the rejection handler',
+      );
+      assert(
+        !messages.some((m: string) => m.includes('timed out after')),
```
charliesantos, line 1406: This test confirms the sip.js rejection ran before the outer timeout, but doesn't verify recovery is re-armed. Worth appending `await clock.tickAsync(200); assert.strictEqual(uaStub.reconnect.callCount, 2)` to prove the backoff schedules a second attempt.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -803,6 +864,69 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+    // signal, and slow reconnects were landing unregistered.
+    if (this._wasRegistered) {
+      this._createAndStartRegisterer();
+    }
+
+    // The server never received the orphaned re-INVITE's ICE/DTLS creds, so
+    // Call has to re-trigger. _inFlightIceRestarts entries stay stale so their
+    // eventual rejection is still suppressed.
+    const recovery: string[] = [];
+    this._pendingIceRestartRecovery.forEach((callSid: string) => recovery.push(callSid));
```
charliesantos: This copy loop can be `const recovery = [...this._pendingIceRestartRecovery]; this._pendingIceRestartRecovery.clear();`.

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: Deferring on the SIP side only would mean a consumer sees the 31009 synchronously in PStream mode and asynchronously in SIP mode, for the same invite() call. That's a behavioral fork in the abstraction the branch exists to maintain.

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: Good catch, you're right that _isReconnecting is already false by the time the flush runs, so a failing send just fails and never re-queues.

I dropped the sentence rather than swapping in the splice/re-entrancy rationale, since that one isn't reachable either: a re-queue needs _isReconnecting true and a second flush needs _onTransportConnect, and both only happen via sip.js's onDisconnect/onConnect delegates, which are WebSocket-event driven and can't fire synchronously inside message.send(). The comment now just says what the method does.

lib/twilio/signaling/sipsignalingadapter.ts
kpchoy: Agreed on the simplification, though [...set] doesn't compile here: tsconfig sets no target so it defaults to ES5, and without downlevelIteration TS gives TS2802 on spreading a Set. Went with Array.from(this._pendingIceRestartRecovery) instead, which has no such restriction and matches existing usage in call.ts and audiohelper.ts.

kpchoy (commented):
Since this a feature branch we have not been updating the Changelog for any changes.

charliesantos (approved):

PR #456, opened 2026-08-14
[VBLOCKS-7025] implement release process gha

kpchoy:
**Contributing to Twilio**

## Pull Request Details

https://twilio-engineering.atlassian.net/browse/VBLOCKS-7025

### Description

GHA release

### Before review
* [ ] Ready for review


.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,370 @@
...
+          esac
+
+          # A final release must have a DATED CHANGELOG heading like
+          # "2.18.4 (August 14, 2026)". Requiring a 4-digit year in the heading
+          # rejects the in-progress placeholder "2.18.4 (Work in progress)", which
+          # a bare "^${RELEASE_VERSION} (" prefix would wave through — shipping a
+          # tarball, tag, and docs whose top entry still says "Work in progress".
+          # The year, not the leading capital, is what separates the two ("Work"
+          # is also capitalized). Any reasonable date format satisfies it.
+          if ! grep -qE "^${RELEASE_VERSION} \(.*[0-9]{4}\)" CHANGELOG.md; then
```
charliesantos: This interpolates `$RELEASE_VERSION` as a regex pattern, so the dots in a semver like `2.18.4` match any character. There's no injection risk since it comes via `env:`, but the unescaped dots are technically incorrect. Escape them first: `ESCAPED_VERSION=$(printf '%s' "$RELEASE_VERSION" | sed 's/[.]/\\./g')` and grep on `"^${ESCAPED_VERSION} \(.*[0-9]{4}\)"`.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,370 @@
...
+# Replaces the CircleCI `release`/`development` plans in .release.json.
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_version:
+        description: "Version to release (e.g. 2.19.0 or 2.19.0-rc1)"
+        required: true
+      development_version:
+        description: "Optional: next dev version to commit after the release (e.g. 2.19.1-dev). Leave blank to skip."
```
charliesantos: When `development_version` is blank the dev-commit step is skipped, leaving the branch tip at the built release tree (dist/, es5/, esm/, docs/, constants.ts) — directories that are gitignored in normal development, so the next push to the branch will diff them as mass deletions. Update the description to warn that omitting this leaves the branch at the built release tree unless an immediate manual dev-version bump follows, or emit a workflow notice when `dry_run == false` and this is empty.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,370 @@
...
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      # The checkout fetches no tags, so a local `git tag` would not notice a tag that
+      # already exists on origin.
+      - name: Verify the tag does not already exist
+        run: |
+          set -euo pipefail
+
+          if git ls-remote --exit-code --tags origin "refs/tags/${RELEASE_VERSION}" >/dev/null 2>&1; then
```
charliesantos: `git ls-remote --exit-code` in an `if` condition can't be protected by `set -euo pipefail`, and it exits non-zero for both "tag not found" and network/auth failure — both fall through as "tag does not exist" with all output suppressed by `>/dev/null 2>&1`. Worth capturing the exit code explicitly: `set +e; git ls-remote ...; RC=$?; set -e` then branch on RC=0 (exists), 1 (proceed), other (network error, abort).

.github/workflows/release.yml
```diff
@@ -0,0 +1,164 @@
...
+            echo "::error::docs/ is missing or empty on ${GITHUB_REF_NAME} — nothing to publish."
+            exit 1
+          fi
+
+          git config user.name "twilio-vblocks-ci"
+          git config user.email "svc.vblocks-ci@twilio.com"
+
+          # Replace the published tree wholesale so files deleted since the last
+          # release do not linger on gh-pages.
+          git rm -rqf . >/dev/null 2>&1 || true
```
charliesantos: `git rm -rqf . >/dev/null 2>&1 || true` swallows all errors, so if the rm fails the following `cp -R ../release/docs/. .` overlays new docs onto the old tracked tree and stale files persist on gh-pages with no signal. Worth dropping the `>/dev/null 2>&1 || true`; to tolerate an empty gh-pages, guard on tracked files existing first (e.g. `git ls-files -z | xargs -0 git rm -rf`).

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,370 @@
...
+
+# No default permissions; each job below declares exactly what it needs.
+permissions: {}
+
+jobs:
+  # The only job that installs dependencies. Keep it read-only.
+  build:
+    name: Build release tree
+    runs-on: ubuntu-latest-large
+    timeout-minutes: 60
```
charliesantos: This build job (install + TS build + lint + unit + es5/esm check + pack) should finish in ~10-15 min; a 60-minute timeout delays the failure signal if it stalls. Consider reducing to 20-25 minutes.

tests/unit/check-version-bump.ts
```diff
@@ -0,0 +1,99 @@
+import * as assert from 'assert';
+
+// Plain CommonJS module (scripts/ is not part of the TS build); require avoids
+// the esModuleInterop dance for a default-less module.exports.
+// tslint:disable-next-line
```
charliesantos: A bare `// tslint:disable-next-line` suppresses every rule on the following `require()`; the rest of the codebase uses named suppressions. Consider naming the specific rule: `// tslint:disable-next-line:no-require-imports`.

charliesantos (changes requested):
The new prepare/release split and its security posture look solid, but the old CircleCI release path is still fully active alongside it. The PR description says this replaces the CircleCI release/development plans, yet `.release.json` still defines active `release`/`development`/`publish` plans and `.circleci/config.yml` still has a working `release-workflow`. Both paths can cut a release for the same version from different build environments and produce two npm publishes, and the CircleCI path bypasses the new OIDC/lifecycle-guard controls. Remove or gate the CircleCI release plans in this PR, or add an explicit deprecation notice documenting which path is authoritative.

.github/workflows/prepare-release.yml
kpchoy: Should we make `development_version` required for final `X.Y.Z` releases specifically while leaving it optional for prereleases?

kpchoy (commented):
Went with full removal rather than deprecation, so there's now a single release path. That closes the double-publish window you flagged: GHA is now the only path that can cut or publish a release.

kpchoy:
Process:
- Human dispatches Prepare Release from master, entering release_version (and development_version, required for a final release).
- GHA validates the inputs, builds the SDK, and runs lint + unit + es5/esm checks.
- GHA commits the built release (X.Y.Z) and its tag, plus the -dev bump, onto a new release/X.Y.Z branch (never to master directly, which the ruleset blocks).
- Human opens a PR from release/X.Y.Z into master and merges it with a merge commit (not squash/rebase, which would orphan the tag).
- Human dispatches Release from the X.Y.Z tag.
- GHA verifies the tag is reachable from master, publishes to npm via OIDC, and refreshes the gh-pages docs.

We merge before publishing because Release only publishes a tag that's reachable from master, and the tag isn't on master until the release/X.Y.Z PR is merged.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,440 @@
...
+      # Catches a transposed or stale version number. Runs before the install, so
+      # the script it calls takes no dependencies.
+      - name: Verify the release version moves forward
+        env:
+          RELEASE_VERSION: ${{ inputs.release_version }}
+        run: |
+          set -euo pipefail
+
+          node ./scripts/check-version-bump.js \
+            "$(node -p "require('./package.json').version")" \
```
mhuynh5757: We have `jq` available on the runners, so you don't have to spin up a whole node process with an inline method to read the `package.json` version. In the end, doesn't really matter too much, but feels more precise with `jq`. I'll leave it up to you if you want to use `jq` vs. leave this as-is.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,440 @@
...
+          do
+            if [ -z "$(ls -A "$RELEASE_DIR" 2>/dev/null)" ]; then
+              echo "::error::${RELEASE_DIR}/ is missing or empty after the build."
+              exit 1
+            fi
+          done
+
+      # A publish cannot be undone: this proves the tree builds and shows exactly
+      # what would ship, without writing anything anywhere.
+      - name: Dry run, pack and preview the publish
```
mhuynh5757, line 220: Nice. I will likely add something like this for the RN SDK 👍

.github/workflows/release.yml
```diff
@@ -0,0 +1,167 @@
...
+    runs-on: ubuntu-latest-large
+    timeout-minutes: 20
+    # Both at job level: `environment:` resolves before any step runs, so a
+    # first-step guard would gate on approval first, and the OIDC token only
+    # carries the environment claim when the job itself asks for the environment,
+    # which is what the npm trusted publisher matches on.
+    if: github.repository_owner == 'twilio'
+    environment:
+      name: production
+      url: https://www.npmjs.com/package/@twilio/voice-sdk/v/${{ github.ref_name }}
```
mhuynh5757, line 39: Could you elaborate on this one? Not sure what it's for, maybe the RN SDK needs this?

.github/workflows/release.yml
```diff
@@ -0,0 +1,167 @@
...
+
+      # Full commit history for the reachability check below; blob:none skips the
+      # built dist/, es5/, esm/ and docs/ trees committed on every past release tag.
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          fetch-depth: 0
+          filter: blob:none
+
+      # The check above constrains the tag's name, not where it came from. voice-js
+      # releases only from master.
```
mhuynh5757, line 64: Probably not something we want to address now, but this won't be true if we ever need to publish a backport. The RN SDK pipeline handles backports if you need a reference. Historically speaking, I don't know if we've ever needed to release a backport i.e. latest is `1.7.0` and we need to release a security patch on `1.6.3`

.github/workflows/release.yml
```diff
@@ -0,0 +1,167 @@
...
+            exit 1
+          fi
+
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          # npm >= 11.5.1, bundled with Node 24, is required for OIDC trusted
+          # publishing.
+          node-version: 24
+
+      # Used over a plain `npm publish` for two guarantees: the tag must match
```
mhuynh5757, line 103: I took a peek at how the platform team's reusable publish action works. It doesn't support specific npm publish tags, such as `beta` or `preview`. Something to note for later if we need to publish different non-latest versions (such as backports).

mhuynh5757 (commented):
Seems solid albeit more complex than the RN SDK implementation. I didn't see anything out-of-place with a manual review, but I left a few comments about backporting and tagging features.

.github/workflows/release.yml
kpchoy: Thanks for checking. The publish action only does the latest tag, so no beta/preview/backport for now. This PR only publishes final X.Y.Z releases. If we need non-latest dist-tags later we'd need a different publish path, with the RN SDK as reference.

.github/workflows/release.yml
kpchoy: Good catch. The check is master-only right now, so a backport (e.g. a patch on 1.6.3 while latest is 1.7.0) would fail it. Leaving it as-is since we've never cut a backport; when we do, this is where we'd add the maintenance branch. Not in this PR.

.github/workflows/release.yml
kpchoy: It's the deployment URL GitHub shows for the production environment; it links the deployment to the published npm version page. Just informational, not required to publish. I think it would be beneficial for rnsdk as well!

mhuynh5757 (approved):

luisrivas (approved):
Thanks, @kpchoy!

charliesantos (approved):

PR #461, opened 2026-08-22
Fix unhandled NotSupportedError when deleting speakers without setSinkId

### LINKED ISSUE #458: [BUG] OutputDeviceCollection.delete() raises an unhandled NotSupportedError on browsers without setSinkId
- [x] I have verified that the issue occurs with the latest twilio.js release and is not marked as a known issue in the [CHANGELOG.md](https://github.com/twilio/twilio-voice.js/blob/master/CHANGELOG.md).
 - [x] I reviewed the [Common Issues](https://github.com/twilio/twilio-voice.js/blob/master/COMMON_ISSUES.md) and open GitHub issues and verified that this report represents a potentially new issue.
 - [x] I verified that the [Quickstart](https://github.com/TwilioDevEd/voice-javascript-sdk-quickstart-node) application works in my environment.
 - [x] I am not sharing any [Personally Identifiable Information (PII)](https://www.twilio.com/docs/glossary/what-is-personally-identifiable-information-pii) or sensitive account information (API keys, credentials, etc.) when reporting this issue.

**Code to reproduce the issue:**

```js
window.addEventListener('unhandledrejection', (e) => {
  console.error('unhandled rejection from the SDK:', e.reason);
});

// On Chrome for Android, where HTMLAudioElement.prototype.setSinkId is absent:
const device = new Twilio.Device(token, { logLevel: 1 });
console.log(device.audio.isOutputSelectionSupported); // false

const call = await device.connect({ params: { To: '+15551234567' } });

// Mid-call, disconnect an audio output device — e.g. a Bluetooth or wired
// headset. No application code touches speakerDevices.
```

**Expected behavior:**

On a browser where output selection can never work, losing an output device should be pure bookkeeping. The SDK already knows the capability is missing — `isOutputSelectionSupported` is `false` — so it should not attempt a sink change it knows will be rejected. Example - Chrome on Android.

**Actual behavior:**

```
NotSupportedError: Audio output selection is not supported by this browser
  at AudioHelper._removeLostOutput (@twilio/voice-sdk/esm/twilio/audiohelper.js)
  at OutputDeviceCollection.delete (@twilio/voice-sdk/esm/twilio/outputdevicecollection.js)
  at Device._updateSinkIds (@twilio/voice-sdk/esm/twilio/device.js)
  at Device._updateSpeakerSinkIds (@twilio/voice-sdk/esm/twilio/device.js)
  at Call._setSinkIds (@twilio/voice-sdk/esm/twilio/call.js)
  at PeerConnection._setSinkIds (@twilio/voice-sdk/esm/twilio/rtc/peerconnection.js)
```

`handled: no`, `mechanism: onunhandledrejection`, no application frame anywhere on the stack.

**Source analysis (2.18.3):**

`lib/twilio/outputdevicecollection.ts:35-52`:

```ts
delete(device: MediaDeviceInfo): boolean {
  this._log.debug('.delete', device);
  const wasDeleted: boolean = !!(this._activeDevices.delete(device));

  const defaultDevice: MediaDeviceInfo = this._availableDevices.get('default')
    || Array.from(this._availableDevices.values())[0];

  if (!this._activeDevices.size && defaultDevice) {
    this._activeDevices.add(defaultDevice);
  }

  // Call _beforeChange so that the implementation can react when a device is
  // removed or lost.
  const deviceIds = Array.from(this._activeDevices.values()).map(deviceInfo => deviceInfo.deviceId);

  this._beforeChange(this._name, deviceIds);   // <-- promise discarded
  return !!wasDeleted;
}
```

`delete()` is the only method on this class that ignores `this._isSupported`. `set()` (`:79`) and the other capability-gated paths all short-circuit when it is `false`. But `delete()` unconditionally re-adds the default device and calls `_beforeChange`, which is `Device._updateSinkIds` → `_updateSpeakerSinkIds` → `Call._setSinkIds` → `PeerConnection._setSinkIds`. On a browser without `setSinkId`, `PeerConnection._isSinkSupported` is `false`, so that chain rejects with `NotSupportedError` — and `delete()` returns a `boolean`, so the rejected promise has nowhere to go.

The trigger is entirely internal: `AudioHelper._updateDevices` calls `speakerDevices.delete(lostDevice)` for every output device that disappears on a `devicechange`, which is exactly what the stack above shows.

**Suggested fix:**

```diff
   delete(device: MediaDeviceInfo): boolean {
     this._log.debug('.delete', device);
+    // Sink selection is impossible on this browser: set() always rejects and
+    // _beforeChange would reject with NotSupportedError. Just drop the device.
+    if (!this._isSupported) {
+      return !!(this._activeDevices.delete(device));
+    }
     const wasDeleted: boolean = !!(this._activeDevices.delete(device));
```

This matches how `set()` and the rest of the class already treat `_isSupported`, and it changes nothing on browsers that do support output selection.

**Software versions:**

- [x] Browser(s): Chrome Mobile 150
- [x] Operating System: Android 10
- [x] twilio.js: 2.18.3
- [x] Third-party libraries (e.g., Angular, React, etc.): React 18, Next.js

Hashim1999164:
## Summary
On browsers without setSinkId support, OutputDeviceCollection.delete still called beforeChange. That path rejects with NotSupportedError, so removing a speaker from an unsupported collection left an unhandled rejection (see #458).

This change removes the device from the active set and returns early when the collection is unsupported, matching how get already behaves. A unit test covers the unsupported delete path.

## Test plan
1. Build constants, then run the outputdevicecollection mocha file with the TypeScript register loader.
2. Confirm the new unsupported delete case passes and existing supported delete cases still pass.

Fixes #458

### DIFF
--- a/lib/twilio/outputdevicecollection.ts
+++ b/lib/twilio/outputdevicecollection.ts
@@ -34,6 +34,13 @@ export default class OutputDeviceCollection {
    */
   delete(device: MediaDeviceInfo): boolean {
     this._log.debug('.delete', device);
+
+    // Sink selection is impossible on this browser: set() always rejects and
+    // _beforeChange would reject with NotSupportedError. Just drop the device.
+    if (!this._isSupported) {
+      return !!(this._activeDevices.delete(device));
+    }
+
     const wasDeleted: boolean = !!(this._activeDevices.delete(device));
 
     const defaultDevice: MediaDeviceInfo = this._availableDevices.get('default')
--- a/lib/twilio/outputdevicecollection.ts
+++ b/lib/twilio/outputdevicecollection.ts
@@ -35,12 +35,6 @@ export default class OutputDeviceCollection {
   delete(device: MediaDeviceInfo): boolean {
     this._log.debug('.delete', device);
 
-    // Sink selection is impossible on this browser: set() always rejects and
-    // _beforeChange would reject with NotSupportedError. Just drop the device.
-    if (!this._isSupported) {
-      return !!(this._activeDevices.delete(device));
-    }
-
     const wasDeleted: boolean = !!(this._activeDevices.delete(device));
 
     const defaultDevice: MediaDeviceInfo = this._availableDevices.get('default')
@@ -48,6 +42,13 @@ export default class OutputDeviceCollection {
 
     if (!this._activeDevices.size && defaultDevice) {
       this._activeDevices.add(defaultDevice);
+    }
+
+    // Sink selection is impossible on this browser: set() always rejects and
+    // _beforeChange would reject with NotSupportedError. Keep the empty-set
+    // default restore above, but skip the implementation callback.
+    if (!this._isSupported) {
+      return wasDeleted;
     }
 
     // Call _beforeChange so that the implementation can react when a device is
--- a/tests/outputdevicecollection.js
+++ b/tests/outputdevicecollection.js
@@ -47,6 +47,35 @@ describe('OutputDeviceCollection', () => {
         assert.equal(collection.get().size, 0);
         assert.equal(onChange.callCount, 0);
       });
+
+      it('should restore the default device without calling _beforeChange', () => {
+        const onChange = sinon.spy(() => Promise.reject(new Error('should not be called')));
+        const deviceDefault = { deviceId: 'default', kind: 'audiooutput' };
+        const fakeDevice = { deviceId: 'lost' };
+        collection = new OutputDeviceCollection('foo', new Map([['default', deviceDefault]]), onChange, false);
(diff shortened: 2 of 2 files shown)


r0busta:
Saw this bug fired in prod for me today. The fix seems to have a correct shape.

The only flag: the early return skips "if the set is now empty, re-add the default device" bookkeeping below it. So on unsupported browsers speakerDevices.get() now returns an empty Set where it previously returned {default}.

Hashim1999164:
After delete on unsupported browsers the default device (or first available) is restored into the set. _beforeChange is still skipped so NotSupportedError does not fire.

kpchoy:
Thanks for your contribution. See: https://github.com/twilio/twilio-voice.js/pull/467

PR #463, opened 2026-08-25
[VBLOCKS-7005] fix: remote re-invite, pc handler restore, cachedAnswer leak

kpchoy:
**Contributing to Twilio**

## Pull Request Details

https://twilio-engineering.atlassian.net/browse/VBLOCKS-7005

### Description
https://github.com/twilio-internal/voice-video-sdks-eng-docs/blob/main/investigations/voice/browser/sip-jssdk-action-items.md
- Item 22: added PeerConnection.processOffer() and a _hasNegotiated routing branch so a server re-INVITE renegotiates on the live PC instead of hanging forever
- Item 23: close() restores Call's original pc.onerror/pc.onfailed, plus a _closed guard so post-close calls reject rather than hang
- Item 24: close() clears _cachedAnswer
- Offerless re-INVITE rejects cleanly so sip.js answers 488 instead of stalling

### Before review
* [ ] Ready for review

### DIFF
--- a/lib/twilio/rtc/peerconnection.ts
+++ b/lib/twilio/rtc/peerconnection.ts
@@ -840,23 +840,38 @@ PeerConnection.prototype.processAnswer = function(sdp, onMediaStarted) {
   });
 };
 /**
- * Answers a remote offer on an open connection (SIP re-INVITE). Skips
- * _initializeMediaStream, which no-ops on an open PC and would replace it.
+ * Answers a remote offer on an established dialog (SIP re-INVITE). Deliberately
+ * mirrors answerIncomingCall minus _initializeMediaStream: that returns false
+ * without calling back once status is 'open' (the caller then waits forever),
+ * and while status is still 'connecting' it rebuilds this.version, discarding
+ * the live connection.
  * @private
  */
 PeerConnection.prototype.processOffer = function(sdp, onAnswerReady, onMediaStarted) {
+  const self = this;
+  function onAnswerError(err) {
+    const errMsg = err.message || err;
+    self.onerror({ info: {
+      code: 31000,
+      message: `Error processing offer: ${errMsg}`,
+      twilioError: new MediaErrors.ClientRemoteDescFailed(),
+    } });
+  }
+
+  // Returning silently would hang the caller: SIP.js blocks on the answer,
+  // and onerror is the only channel that reaches it.
   if (this.status === 'closed') {
+    onAnswerError(new Error('Connection is closed'));
     return;
   }
 
   const processedSdp = this._maybeSetIceAggressiveNomination(sdp);
   this._answerSdp = processedSdp;
 
-  const self = this;
   function onAnswerSuccess() {
-    // close() may have landed while createAnswer was in flight, which nulls
-    // version.pc, so getSDP() would throw. Mirrors answerIncomingCall.
+    // close() nulls version.pc, so getSDP() would throw.
     if (self.status === 'closed') {
+      onAnswerError(new Error('Connection closed while creating the answer'));
       return;
     }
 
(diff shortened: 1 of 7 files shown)


lib/twilio/rtc/peerconnection.ts
```diff
@@ -839,6 +839,48 @@ PeerConnection.prototype.processAnswer = function(sdp, onMediaStarted) {
     } });
   });
 };
+/**
+ * Answers a remote offer on an open connection (SIP re-INVITE). Skips
+ * _initializeMediaStream, which no-ops on an open PC and would replace it.
+ * @private
+ */
+PeerConnection.prototype.processOffer = function(sdp, onAnswerReady, onMediaStarted) {
```
charliesantos, line 850: This closed-status guard returns without invoking `onAnswerReady`, `onMediaStarted`, or `onerror`, so `SipSessionDescriptionHandler.setDescription()` - whose only resolve path is the `processOffer` callback and whose only reject path is `PeerConnection.onerror`/`onfailed` firing `_failPending` - hangs forever instead of rejecting. `Call` closes the PeerConnection synchronously in `disconnect()`/`reject()`/`_onCancel()` while SIP.js closes the SDH later at session teardown, so a remote re-INVITE landing in that window takes exactly this path - the same hung-promise class this PR fixes at its other two call sites. Have the closed-status branches here and in `onAnswerSuccess` (and `processAnswer`, which shares the pattern) invoke the error callback so the SDH promise rejects, and add a test that closes the PeerConnection while a `processOffer` is in flight and asserts `setDescription()` rejects.

charliesantos, line 850: This is a near-copy of `answerIncomingCall` below, and diffing the two bodies gives four deltas of which only one is load-bearing: swapping the `_initializeMediaStream()` guard for the `status === 'closed'` guard, which is exactly why a separate entry point is needed. The other three are no-ops - `this.callSid` was already set by the initial exchange, `_setupRTCDtlsTransportListener` is installed on the initial exchange and early-returns when `onstatechange` is already set, and the actpass rewrite never reached the peer connection at all (see the note further down). The test scaffolding got duplicated the same way, in a 106-line `processOffer` context mirroring the existing `answerIncomingCall` and `processAnswer` contexts. Worth extracting a shared `_answerRemoteOffer(sdp, onAnswerReady, onMediaStarted, errorLabel)` with both methods as thin wrappers - or, if copy-paste is the deliberate low-risk call given this file's `@ts-nocheck` prototype style and bind-based tests, saying so in a comment so the next reader doesn't have to re-derive which delta matters.

lib/twilio/signaling/sipsignalingadapter.ts
```diff
@@ -513,9 +502,20 @@ export class SipSignalingAdapter extends EventEmitter implements SignalingAdapte
...
+    // Arm the SDH directly. Invite options stick to the dialog and replay
+    // onto later server-initiated re-INVITEs.
+    const sdh = session.sessionDescriptionHandler as
+      Partial<SipSessionDescriptionHandler> | undefined;
+    if (typeof sdh?.requestIceRestart !== 'function') {
+      this._log.warn('iceRestart: no usable description handler for callSid', callSid);
+      this.emit('hangup', { callsid: callSid });
+      return;
+    }
+    sdh.requestIceRestart();
```
charliesantos, line 514: This arms the SDH's one-shot `_iceRestartRequested` flag before `session.invite()` is known to succeed; if the invite rejects with `RequestPendingError`, the catch below just logs and returns without rolling the flag back. No user-visible failure traced today - the flag is either re-armed by a later legitimate `iceRestart()` or bypassed by `getDescription()`'s cached-answer check - but it does break the flag's documented one-shot contract. Worth arming only once the invite is known to proceed, or resetting the flag in the `RequestPendingError` branch.

lib/twilio/rtc/peerconnection.ts
```diff
@@ -839,6 +839,48 @@ PeerConnection.prototype.processAnswer = function(sdp, onMediaStarted) {
...
+  function onAnswerError(err) {
+    const errMsg = err.message || err;
+    self.onerror({ info: {
+      code: 31000,
+      message: `Error processing offer: ${errMsg}`,
+      twilioError: new MediaErrors.ClientRemoteDescFailed(),
+    } });
+  }
+
+  // NOTE(VBLOCKS-7005): no a=setup:actpass -> passive rewrite here. The DTLS
```
charliesantos: This rationale doesn't hold. In `answerIncomingCall` the actpass rewrite is assigned only to `this._answerSdp` (line 889), while `version.processSDP` receives the un-rewritten `sdp` (line 910) - and `_answerSdp` is written in three places and read nowhere outside test assertions. The rewrite never reached the peer connection or the wire, so it cannot have fixed the DTLS role on the initial exchange. Skipping it here is still correct; worth rewording the note to the actual reason (the field is diagnostic-only and unread, so the rewrite was already a no-op) so a future reader doesn't trust the DTLS claim.

charliesantos (changes requested):
No blockers - one correctness gap and two cleanups flagged inline.

CHANGELOG.md isn't updated even though this fixes three user-observable call-hanging bugs; worth an entry covering the remote re-INVITE and close() fixes.

lib/twilio/rtc/peerconnection.ts
kpchoy: I updated the comment. I think we should leave as is, because extracting a shared `_answerRemoteOffer` would have meant editing answerIncomingCall, a released VSP path, to serve an unreleased SIP one.

kpchoy:
Also since this is a feature branch there is no need for changelog

charliesantos (approved):

PR #467, opened 2026-08-26
[VBLOCKS-7064] fix: output device unhandled promise rejection

### LINKED ISSUE #458: [BUG] OutputDeviceCollection.delete() raises an unhandled NotSupportedError on browsers without setSinkId

kpchoy:
**Contributing to Twilio**

## Pull Request Details

https://twilio-engineering.atlassian.net/browse/VBLOCKS-7064

### Description

Fixes: https://github.com/twilio/twilio-voice.js/issues/458

### Before review
* [ ] Ready for review

### DIFF
--- a/lib/twilio/outputdevicecollection.ts
+++ b/lib/twilio/outputdevicecollection.ts
@@ -28,13 +28,19 @@ export default class OutputDeviceCollection {
   /**
    * Delete a device from the collection. If no devices remain, the 'default'
    * device will be added as the sole device. If no `default` device exists,
-   * the first available device will be used.
+   * the first available device will be used. On browsers without output
+   * selection, the collection is left empty.
    * @param device - The device to delete from the collection
    * @returns whether the device was present before it was deleted
    */
   delete(device: MediaDeviceInfo): boolean {
     this._log.debug('.delete', device);
     const wasDeleted: boolean = !!(this._activeDevices.delete(device));
+
+    // Without sink support _beforeChange rejects, and nothing can catch it.
+    if (!this._isSupported) {
+      return wasDeleted;
+    }
 
     const defaultDevice: MediaDeviceInfo = this._availableDevices.get('default')
       || Array.from(this._availableDevices.values())[0];
@@ -47,8 +53,15 @@ export default class OutputDeviceCollection {
     // removed or lost.
     const deviceIds = Array.from(this._activeDevices.values()).map(deviceInfo => deviceInfo.deviceId);
 
-    this._beforeChange(this._name, deviceIds);
-    return !!wasDeleted;
+    // Nothing can await a synchronous method. The wrapper also tolerates a
+    // _beforeChange that returns a non-promise or throws.
+    new Promise(resolve => {
+      resolve(this._beforeChange(this._name, deviceIds));
+    }).catch(reason => {
+      this._log.warn(`Unable to update audio output devices. ${reason}`);
+    });
+
+    return wasDeleted;
   }
 
   /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 :warning: **Important**: If you are upgrading to version 2.3.0 or later and have firewall rules or network configuration that blocks any unknown traffic by default, you need to update your configuration to allow connections to the new DNS names and IP addresses. Please refer to this [changelog](#230-january-23-2023) for more details.
 
+2.18.5 (Work in progress)
+=========================
+
+Bug Fixes
+---------
+
+- Fixed an [issue](https://github.com/twilio/twilio-voice.js/issues/458) where losing an audio output device during a call caused an unhandled promise rejection. On browsers without audio output selection, such as Chrome on Android, `OutputDeviceCollection.delete()` attempted a sink change that can never apply there, raising `NotSupportedError` without affecting call audio; it no longer attempts one. On browsers that do support selection, a failed switch to the fallback device is now logged as a warning. Thanks @ifabijanovic for reporting this.
+
+  Note: on browsers without audio output selection, `device.audio.speakerDevices.get()` and `ringtoneDevices.get()` now stay empty after a device is lost instead of reporting `default`.
+
 2.18.4 (August 31, 2026)
 ========================
 
--- a/tests/audiohelper.js
+++ b/tests/audiohelper.js
@@ -1123,6 +1123,26 @@ describe('AudioHelper', () => {
       });
     });
 
+    context('when output selection is not supported', () => {
+      beforeEach(async () => {
+        onActiveOutputsChanged = sinon.spy(() => Promise.resolve());
+        audio = new AudioHelper(onActiveOutputsChanged, null, {
+          getUserMedia,
+          mediaDevices,
+          setSinkId: undefined,
+        });
+        await new Promise(resolve => setTimeout(resolve));
+      });
+
+      it('should not trigger onActiveOutputsChanged when an output device is lost', async () => {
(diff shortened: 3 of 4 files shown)


lib/twilio/outputdevicecollection.ts
```diff
@@ -28,14 +28,20 @@ export default class OutputDeviceCollection {
...
    * @returns whether the device was present before it was deleted
    */
   delete(device: MediaDeviceInfo): boolean {
     this._log.debug('.delete', device);
     const wasDeleted: boolean = !!(this._activeDevices.delete(device));
 
+    // Without sink support _beforeChange rejects, and nothing can catch it.
+    if (!this._isSupported) {
+      return wasDeleted;
+    }
```
mhuynh5757, line 43: If we return a value here, does that imply that some "action" or side-effect actually occurred? Is it more accurate to throw here? I guess what I'm saying is that the whole method didn't really invoke, so not sure if early return here without an error suggests that something happened when it really didn't, in this case `_beforeChange`. But it also seems like this is intentional, and that `_beforeChange` _shouldn't_ be invoked when `this._isSupported` is false.

lib/twilio/outputdevicecollection.ts
```diff
@@ -47,8 +53,15 @@ export default class OutputDeviceCollection {
...
-    return !!wasDeleted;
+    // Nothing can await a synchronous method. The wrapper also tolerates a
+    // _beforeChange that returns a non-promise or throws.
+    new Promise(resolve => {
+      resolve(this._beforeChange(this._name, deviceIds));
+    }).catch(reason => {
+      this._log.warn(`Unable to update audio output devices. ${reason}`);
+    });
+
+    return wasDeleted;
```
mhuynh5757, line 64: The `!!` is a good removal, but let's just leave it in IMO. Less overall changes so less risk of breaking something.

mhuynh5757 (changes requested):

lib/twilio/outputdevicecollection.ts
kpchoy: - The return value's contract is "whether the device was present before it was deleted", not "whether `_beforeChange` ran".
- Throwing would recreate this exact bug. The only production caller is `AudioHelper._removeLostOutput`, synchronous inside `_updateDevices`, inside `_updateAvailableDevices`' `.then()`. A throw there rejects that promise, which is unhandled at `audiohelper.ts:320` and `:804`, so we'd swap one unhandled rejection for another.
- It would also change documented runtime behavior. `delete()` is public and returns `false` there today, so anyone calling `speakerDevices.delete()` directly would start getting an exception instead.

lib/twilio/outputdevicecollection.ts
kpchoy: I want to keep this one for hygiene. `wasDeleted` is already a `boolean`, so putting `!!` back here means returning the same variable two different ways in one function. 
Line 38:
> const wasDeleted: boolean = !!(this._activeDevices.delete(device));

mhuynh5757 (dismissed):

mhuynh5757 (approved):

PR #470, opened 2026-09-02
[VBLOCKS-7102] fix: register retry

kpchoy:
**Contributing to Twilio**

## Pull Request Details

https://twilio-engineering.atlassian.net/browse/VBLOCKS-7102

### Description

Fix reconnect race surfaces false 20101 AccessTokenInvalid after transport drop

### Before review
* [ ] Ready for review

### DIFF
--- a/lib/twilio/pstream.ts
+++ b/lib/twilio/pstream.ts
@@ -214,7 +214,9 @@ PStream.prototype.sendMessage = function(
 
 PStream.prototype.register = function(mediaCapabilities) {
   const regPayload = { media: mediaCapabilities };
-  this._publish('register', regPayload, true);
+  // A queued register is replayed alongside `listen`, before the server
+  // validates the token, and earns a 31204. Re-register on `connected` covers it.
+  this._publish('register', regPayload, false);
 };
 
 PStream.prototype.invite = function(sdp, callsid, params) {
--- a/tests/unit/device.ts
+++ b/tests/unit/device.ts
@@ -815,6 +815,23 @@ describe('Device', function() {
           sinon.assert.calledOnce(spy);
         });
 
+        it('should not reject when the stream goes offline before the re-register completes', async () => {
+          await registerDevice();
+
+          const warnStub = device['_log'].warn = sinon.stub();
+          pstream.emit('offline');
+          await clock.tickAsync(0);
+          pstream.emit('connected', { region: 'EU_IRELAND' });
+          await clock.tickAsync(0);
+
+          // A second drop before the server `ready` rejects the re-register.
+          pstream.emit('offline');
+          await clock.tickAsync(0);
+
+          sinon.assert.calledOnce(warnStub);
+          assert.equal(device.state, Device.State.Unregistered);
+        });
+
         it('should update the preferred uri', () => {
           pstream.emit('connected', { region: 'EU_IRELAND', edge: Edge.Dublin });
           assert.equal(device['_preferredURI'], ['wss://voice-js.dublin.twilio.com/signal']);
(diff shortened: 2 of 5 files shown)


lib/twilio/pstream.ts
```diff
@@ -214,7 +214,9 @@ PStream.prototype.sendMessage = function(
 
 PStream.prototype.register = function(mediaCapabilities) {
   const regPayload = { media: mediaCapabilities };
-  this._publish('register', regPayload, true);
+  // A queued register is replayed alongside `listen`, before the server
+  // validates the token, and earns a 31204. Re-register on `connected` covers it.
+  this._publish('register', regPayload, false);
```
mhuynh5757, line 219: The third value is `shouldRetry` yeah? Does this break any flow outside of the signaling-reconnect scenario?

mhuynh5757 (commented):
Had a question about the design before approving/requesting changes.

lib/twilio/pstream.ts
kpchoy: @mhuynh5757 I am inclined to say no since all the unit and e2e pass. However if you have a test case you think I am missing let me know.

mhuynh5757: Thinking maybe a simple registration under unstable network conditions:

```
Alice loads Voice JS Softphone
Alice is on unstable internet connection
Alice attempts registration
Internet connection goes in and out
```

Before this change: Alice's SDK auto-attempts reconnection
After this change: Alice's SDK gives up after the first ws failure

But maybe that doesn't matter, since websocket reconnection fires the `connected` event like your comment states. WDYT? Worth trying it?

kpchoy: 👍  Added a unit test for this scenario, please take a look

tests/unit/device.ts
```diff
@@ -815,6 +815,26 @@ describe('Device', function() {
           sinon.assert.calledOnce(spy);
         });
 
+        it('should attempt a re-register if the stream dropped while registering', async () => {
+          // Never send `ready`, so the device is still registering when the
+          // stream drops.
+          const registerPromise = device.register();
+          const didReject = registerPromise.then(() => false, () => true);
```
mhuynh5757: Very minor nitpick: `registerPromise.then(() => false).catch(() => true)`

Feels more readable. Very negligible nitpick though.

tests/unit/device.ts
```diff
@@ -815,6 +815,26 @@ describe('Device', function() {
...
+          await clock.tickAsync(0);
+          assert.equal(device.state, Device.State.Registering);
+
+          pstream.register.resetHistory();
+          pstream.emit('offline');
+          await clock.tickAsync(0);
+          assert.equal(await didReject, true);
+          assert.equal(device.state, Device.State.Unregistered);
+
+          pstream.emit('connected', { region: 'EU_IRELAND' });
```
mhuynh5757, line 832: IIRC, this works because it fires the `shouldReRegister` logic right? Just asking for my personal understanding.

mhuynh5757 (dismissed):

tests/unit/device.ts
kpchoy: Yes, exactly. The `offline` emit two lines up runs `_onSignalingOffline`, which does:

```ts
this._shouldReRegister = this.state !== Device.State.Unregistered;
```

State is `Registering` at that point, so it lands on `true`. Then `connected` runs `_onSignalingConnected`, which sees `_shouldReRegister` and calls `register()`, and that is the `pstream.register` the assertion counts.

The `!== Unregistered` rather than `=== Registered` is the whole reason your Alice scenario recovers, so I checked that the test actually pins it: narrowing it to `=== Device.State.Registered` makes this test fail while the other four re-register tests still pass.

mhuynh5757 (approved):

